PR #66, opened 2026-05-09
Lint .coderabbit.yaml instruction fields with content rules

### LINKED ISSUE #64: Lint .coderabbit.yaml with content rules
## Summary

`.coderabbit.yaml` review instructions are consumed by an LLM, making them a natural target for skillsaw's content intelligence rules (weak language, critical positioning, banned references, etc.).

## Context

CodeRabbit config files can contain `instructions` fields at various levels (global, per-path, per-tool) that guide how the AI reviewer behaves. These are LLM prompts — the same content quality issues that affect SKILL.md and CLAUDE.md apply here.

## Desired behavior

- Detect `.coderabbit.yaml` at repo root
- Extract instruction text fields from the config
- Run content rules against them (weak-language, critical-position, tone-consistency, etc.)
- New format flag (e.g. `HAS_CODERABBIT`) 

## References

- [CodeRabbit config schema](https://docs.coderabbit.ai/reference/configuration)

not-stbenjam:
## Summary

- Add `coderabbit-yaml-valid` rule to validate `.coderabbit.yaml` YAML syntax (severity: error, auto-enabled)
- Add `coderabbit-instructions` rule to check instruction text fields for content quality issues like weak/hedge language (severity: warning, auto-enabled)
- Extract and lint instruction text from `reviews.instructions`, `reviews.path_instructions[].instructions`, `reviews.tools.<tool>.instructions`, and `chat.instructions`
- Line numbers are reported for violations when possible

Fixes #64

## Test plan

- [x] `make test` -- 495 tests pass
- [x] `make lint` -- formatting clean
- [x] `make update` -- generated files regenerated
- [x] `python -m skillsaw .` -- no false positives on skillsaw repo (which has a `.coderabbit.yaml`)
- [x] Test against `openshift-eng/ai-helpers` -- exits 0

## Summary by CodeRabbit

* **New Features**
  * Added two auto-enabled CodeRabbit checks for `.coderabbit.yaml`: YAML validity (error) and instruction content-quality scanning (warning).
  * CLI: allow explicitly overriding detected repository type via a repeatable option.
  * Repo detection now supports multiple simultaneous types and uses that set across discovery and linting.
  * Formatter: verbose JSON output includes the list of detected repo types.

* **Documentation**
  * README and examples updated to document the new CodeRabbit rule group.

* **Tests**
  * Added comprehensive tests for CodeRabbit rules, multi-type detection, and defaults.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/66)

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -1,4 +1,4 @@
-# skillsaw
+# agent-a2c9f10f768c5389e
 
 ## Skills
 
@@ -60,6 +60,44 @@ Claude Code plugins, and plugin marketplaces.
 Rules MUST report line numbers in violations whenever the violation can be
 traced to a specific line in a file. This enables the GitHub Action to post
 inli
+
+## Skills
+
+### skillsaw-ecosystem-scout
+
+Survey the AI coding assistant and agentic tool ecosystem, assess skillsaw's competitive position, identify emerging patterns and missing capabilities, and produce a prioritized strategic report as a GitHub issue.
+
+**Compatibility:** Requires git, gh CLI, and internet access (WebFetch, WebSearch)
+
+### skillsaw-issue-solver
+
+Pick up open GitHub issues from collaborators and create PRs to solve them. Use when triaging and resolving reported bugs or feature requests.
+
+**Compatibility:** Requires git, gh CLI, and internet access
+
+### skillsaw-maintenance
+
+Analyze upstream specs (agentskills.io, Claude Code plugin/marketplace format) for changes, identify gaps in skillsaw's rule coverage, and create or update PRs to close those gaps. Use when performing periodic maintenance on the skillsaw linter.
+
+**Compatibility:** Requires git, gh CLI, and internet access
+
+### skillsaw-pr-review
+
+Review open PRs in skillsaw, fix failing CI, address reviewer feedback, and push updates. Use for triaging and fixing up existing pull requests.
+
+**Compatibility:** Requires git, gh CLI, and internet access
+
+### skillsaw-release
+
+Bump the version, generate release notes, and publish a new skillsaw release to PyPI. Use when cutting a new release.
+
+**Compatibility:** Requires git, gh CLI, and internet access
+
+### skillsaw-review-panel
+
+Serial multi-specialist code review panel for skillsaw PRs. Runs 5 specialist reviewers (Architecture, Python Expert, Security & Supply Chain, QA Engineer, Technical Writer) inline then synthesizes a single verdict.
+
+**Compatibility:** Requires git, gh CLI, and internet access
 
 
 ---
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -1,5 +1,3 @@
-# agent-a2c9f10f768c5389e
-
 ## Skills
 
 ### skillsaw-ecosystem-scout
--- a/.skillsaw.yaml.example
+++ b/.skillsaw.yaml.example
@@ -209,6 +209,16 @@ rules:
     enabled: auto
     severity: info
 
+  # .coderabbit.yaml must be valid YAML
+  coderabbit-yaml-valid:
+    enabled: auto
+    severity: error
+
+  # Check .coderabbit.yaml instruction fields for content quality issues
+  coderabbit-instructions:
+    enabled: auto
+    severity: warning
+
 # Load custom rules from these files
 custom-rules: []
 
--- a/.skillsaw.yaml.example
+++ b/.skillsaw.yaml.example
@@ -214,11 +214,6 @@ rules:
     enabled: auto
     severity: error
 
-  # Check .coderabbit.yaml instruction fields for content quality issues
-  coderabbit-instructions:
-    enabled: auto
-    severity: warning
-
 # Load custom rules from these files
 custom-rules: []
 
--- a/README.md
+++ b/README.md
@@ -428,12 +428,11 @@ Rules that go beyond structural validation to analyze the *quality* of instructi
 
 ### CodeRabbit
 
-Validates `.coderabbit.yaml` config files. Checks YAML syntax and scans instruction text fields (`reviews.instructions`, per-path instructions, per-tool instructions, `chat.instructions`) for content quality issues such as weak/hedge language. Auto-enabled when `.coderabbit.yaml` is detected.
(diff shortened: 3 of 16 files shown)


not-stbenjam:
## Review Panel Verdict

**Disposition**: APPROVE (with suggestions)

---

### Specialist Findings

**Architecture Reviewer**

- **SUGGESTION** — `src/skillsaw/rules/builtin/coderabbit.py:24-34` — **Duplicated weak language patterns.** The `_WEAK_LANGUAGE_PATTERNS` list duplicates patterns from `content_analysis.py`'s `_HEDGING` list (e.g., `try to`, `perhaps`, `if possible`). Consider extracting shared patterns into `utils.py` or importing from `content_analysis.py`. This avoids the lists drifting over time. Not blocking because the coderabbit rule operates on a different file type (YAML instruction fields vs. markdown files) and may intentionally want a simpler subset.

- **NOTE** — `src/skillsaw/rules/builtin/coderabbit.py:274` — **Direct `RuleViolation` construction bypasses configured severity.** `_check_weak_language` creates `RuleViolation(severity=Severity.WARNING, ...)` directly instead of using `self.violation()`. This means if a user overrides the severity in their config (e.g., `severity: error`), the override is ignored. The `CoderabbitYamlValidRule` class correctly uses `self.violation()`, so this is inconsistent within the same file.
  - **Recommended action:** Change `_check_weak_language` to accept a reference to the rule instance and use `self.violation()`, or pass the configured severity as a parameter.

- **NOTE** — Auto-enable behavior: Both rules set `repo_types = None` and `formats = None` (inherited defaults), so with `enabled: "auto"`, `is_rule_enabled()` returns `True` unconditionally for all repos. The rules themselves bail out when `.coderabbit.yaml` is absent, so this is functionally correct but differs from the pattern used by content rules (which set `formats = ALL_INSTRUCTION_FORMATS`). Acceptable as-is since the early return is clean.

**Python Expert**

- **NOTE** — `src/skillsaw/rules/builtin/coderabbit.py:14` — **Unused import.** `Dict` is imported from `typing` but never used in the module. Only `Any`, `List`, `Optional`, and `Tuple` are used.

- **SUGGESTION** — `src/skillsaw/rules/builtin/coderabbit.py:30` — **Overly broad `consider` pattern.** The pattern `\bconsider\b` will match legitimate uses such as "we don't consider this a bug" or "consider the following constraints." The existing `content_analysis.py` uses the more precise `\bconsider\s+(?:using|adding|implementing|creating|moving|switching|enabling)\b`. Recommend tightening this pattern to reduce false positives.

- **NOTE** — `src/skillsaw/rules/builtin/coderabbit.py:115` — `_find_nth_key_line` for path_instructions uses `len(results)` as the occurrence index. This coupling between the accumulator state and the line-finder is fragile. If a new instruction source is added before `path_instructions` processing, the line numbers could be wrong. Works correctly today, but worth a comment explaining the assumption.

- **NOTE** — The code is compatible with Python 3.9+ (uses `from __future__ import annotations`, `re.compile`, standard typing). No walrus operators or match statements. Good.

**Security & Supply Chain Reviewer**

No issues found.

- Uses `yaml.safe_load` (not `yaml.load`) — no deserialization risk.
- No `subprocess`, no `shell=True`, no `eval`/`exec`.
- No new dependencies added — `yaml` (PyYAML) is already a project dependency.
- File paths are constructed from `context.root_path` — no path traversal risk.
- Version bump to 0.6.1 in `pyproject.toml` and `__init__.py` is consistent.

**QA Engineer**

- **NOTE** — Test coverage is strong: 31 tests covering both rules, the extraction helper, config defaults, edge cases (empty YAML, invalid YAML, non-mapping top-level, unreadable file, non-string instructions), and line number reporting. Good use of the project's existing `temp_dir` fixture from `conftest.py`.

- **SUGGESTION** — **Missing test: truncation of weak-language phrases.** The `_check_weak_language` method has logic to show only 3 phrases and append "(and N more)". No test exercises this path. Suggest adding a test with 4+ weak language patterns in a single instruction field.

  Example input: `"Maybe try to do this, perhaps if possible, it would be nice to consider it when feasible."`

  Expected: violation message contains "(and" indicating truncation.

- **SUGGESTION** — **Missing test: `consider` false positives.** Given the overly broad `\bconsider\b` pattern, a test demonstrating it flags "consider the following" would document the known false-positive behavior and serve as a regression baseline if the pattern is later tightened.

- **NOTE** — Tests import `_extract_instructions` (a private function) directly, which is fine for unit testing internals but should be noted as a test coupling point.

**Technical Writer**

- **NOTE** — README, `.skillsaw.yaml.example`, and `generate-docs.py` are all updated consistently. The CodeRabbit section appears in the correct position.

- **NOTE** — The group description says "Auto-enabled when `.coderabbit.yaml` is detected," but the actual mechanism is that the rules are always enabled (since `repo_types` and `formats` are both `None`) and silently skip when the file is absent. The functional behavior is equivalent, but the description implies a detection step that doesn't exist. Minor inaccuracy — not blocking.

- **NOTE** — The generated README section is missing the "Autofix" column that `generate-docs.py` would produce. This suggests `make update` was run, but the script output differs from the hand-written table. However, the main branch also lacks this column, so this is a pre-existing discrepancy, not introduced by this PR.

---

### Panel Synthesis

This is a clean, well-structured addition of two new lint rules following established codebase patterns. The implementation is sound, the test coverage is thorough (31 tests), and the documentation is consistent. No blocking issues were found.

The most substantive suggestion is the overly broad `\bconsider\b` regex pattern which will produce false positives on legitimate uses of the word. The existing content analysis module has a more refined pattern that could serve as a model. The direct `RuleViolation` construction in `_check_weak_language` (bypassing configured severity) is a real correctness issue but is low-impact since most users won't override the severity.

No security, compatibility, or architectural concerns. The change is additive-only, defaults to `enabled: auto`, and cannot break existing users.

---

### Required Actions Before Merge

None. No blocking findings.

---

### Optional Follow-ups

1. **Tighten the `consider` pattern** to `\bconsider\s+(?:using|adding|implementing|...)\b` to match the precision level of `content_analysis.py` and avoid false positives.
2. **Use `self.violation()` in `_check_weak_language`** instead of constructing `RuleViolation` directly, so that user severity overrides are respected.
3. **Remove unused `Dict` import** from `typing`.
4. **Add a test for the truncation path** (4+ weak language phrases showing "(and N more)").
5. **(Optional)** Consider extracting shared weak-language patterns to avoid drift between `coderabbit.py` and `content_analysis.py`.

---

<sub>Generated by [skillsaw-review-panel](https://github.com/stbenjam/skillsaw/tree/main/.apm/skills/skillsaw-review-panel)</sub>

PR #82, opened 2026-05-10
fix: --type flag now drives discovery, not just rule selection

not-stbenjam:
## Summary

- Added `repo_type` parameter to `RepositoryContext.__init__` so discovery (`_discover_plugins`, `_discover_skills`, etc.) runs using the overridden type instead of the auto-detected one
- Added `--type` CLI flag to the `lint` subcommand, accepting any valid `RepositoryType` value (single-plugin, marketplace, agentskills, dot-claude)
- When `--type` is not provided, behavior is identical to before (auto-detection)

**Bug:** Previously, overriding `repo_type` after the constructor would only control which rules fire but not what data those rules see. Plugins/skills/marketplace data would still reflect the auto-detected type, producing misleading results.

## Test plan

- [x] Added `test_repo_type_override_marketplace` -- override changes discovery behavior
- [x] Added `test_repo_type_override_forces_discovery` -- core regression test for the bug
- [x] Added `test_repo_type_override_to_dot_claude` -- override to UNKNOWN finds no plugins
- [x] Added `test_repo_type_override_none_uses_autodetect` -- None is identical to auto-detect
- [x] All 468 tests pass (`make test`)
- [x] Lint clean (`make lint`)
- [x] Tested against openshift-eng/ai-helpers -- exit 0

## Version 0.6.1

* **New Features**
  * Added `--type` option to the `lint` command, enabling users to manually override the automatic repository type detection. This allows you to explicitly specify whether a repository should be treated as a single plugin, marketplace, or Claude workspace, regardless of its directory structure.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/82)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "skillsaw"
-version = "0.6.0"
+version = "0.6.1"
 description = "A configurable linter for agent skills, plugins, and AI coding assistant context"
 readme = "README.md"
 authors = [
--- a/src/skillsaw/__init__.py
+++ b/src/skillsaw/__init__.py
@@ -2,7 +2,7 @@
 skillsaw - A configurable linter for agent skills, plugins, and AI coding assistant context
 """
 
-__version__ = "0.6.0"
+__version__ = "0.6.1"
 
 from .rule import Rule, RuleViolation, Severity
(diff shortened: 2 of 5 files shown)


not-stbenjam:
Re-doing: rebasing on upstream/main, no version bump.

PR #89, opened 2026-05-10
fix: pass since_version for custom rules to enforce version gate

not-stbenjam:
## Summary
- Custom rules loaded via `--custom-rules` omitted the `since_version` parameter when calling `is_rule_enabled()`, defaulting to `"0.1.0"`. This allowed a custom rule declaring e.g. `since = "0.8.0"` to fire even when the config had `version: "0.7.0"`, bypassing the version gate.
- Added `since_version=rule_instance.since` to the `is_rule_enabled` call in `_load_custom_rule()`, matching the existing pattern in `_load_builtin_rules()`.
- Added a test (`test_custom_rule_version_gate`) that verifies a custom rule with `since = "0.8.0"` is excluded when config version is `"0.7.0"` and included when config version is `"0.8.0"`.

## Test plan
- [x] `make test` passes (821 passed, 14 skipped)
- [x] `make lint` passes
- [x] `make update` ran successfully
- [x] New test `test_custom_rule_version_gate` covers the exact bug scenario

### DIFF
--- a/src/skillsaw/linter.py
+++ b/src/skillsaw/linter.py
@@ -127,6 +127,7 @@ class Linter:
                     self.context,
                     rule_instance.repo_types,
                     rule_instance.formats,
+                    since_version=rule_instance.since,
                 ):
                     self.rules.append(rule_instance)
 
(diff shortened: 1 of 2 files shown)


stbenjam:
Custom rules are explicitly listed in .skillsaw.yaml, so this is rejected

PR #93, opened 2026-05-10
fix: reject booleans in numeric type checks

not-stbenjam:
## Summary

- Add `isinstance(val, bool)` exclusion to `stripComponents` validation in `openclaw.py` and eval `id` validation in `agentskills.py`
- Python's `bool` is a subclass of `int`, so `isinstance(True, (int, float))` returns `True`, allowing `stripComponents: true` and `id: true` to silently pass numeric validation
- Added tests verifying that boolean values are correctly rejected in both checks

## Test plan

- [x] `test_install_strip_components_boolean_fails` verifies `stripComponents: true` is flagged
- [x] `test_evals_entry_boolean_id_fails` verifies `id: true` is flagged
- [x] Full test suite passes (822 passed)
- [x] Lint passes

### DIFF
--- a/src/skillsaw/rules/builtin/openclaw.py
+++ b/src/skillsaw/rules/builtin/openclaw.py
@@ -321,8 +321,9 @@ class OpenclawMetadataRule(Rule):
                     )
                 )
 
-            if "stripComponents" in entry and not isinstance(
-                entry["stripComponents"], (int, float)
+            if "stripComponents" in entry and (
+                not isinstance(entry["stripComponents"], (int, float))
+                or isinstance(entry["stripComponents"], bool)
             ):
                 violations.append(
                     self.violation(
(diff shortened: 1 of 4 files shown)


src/skillsaw/rules/builtin/openclaw.py
```diff
@@ -321,8 +321,9 @@ def _check_install(self, openclaw, skill_md, line_map, violations):
                     )
                 )
 
-            if "stripComponents" in entry and not isinstance(
-                entry["stripComponents"], (int, float)
+            if "stripComponents" in entry and (
+                not isinstance(entry["stripComponents"], (int, float))
+                or isinstance(entry["stripComponents"], bool)
             ):
```
stbenjam, line 327: bool makes sense to me here, rejecting

PR #94, opened 2026-05-10
fix: --max-iterations 0 silently ignored due to truthy check

not-stbenjam:
## Summary

- Fixed `if args.max_iterations:` in `__main__.py` line 533 which evaluates to `False` when `--max-iterations 0` is passed, silently ignoring the override
- Changed to `if args.max_iterations is not None:` so that zero is correctly applied
- Added regression tests covering zero, positive, and omitted values

## Test plan

- [x] `make test` passes (823 passed, 14 skipped)
- [x] `make lint` passes
- [x] `make update` passes with no generated file changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `--max-iterations` CLI flag to correctly accept and apply a value of `0` instead of only truthy values.

* **Tests**
  * Added comprehensive test coverage for `--max-iterations` CLI override behavior, including edge cases and scenarios when the flag is omitted.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/94)

### DIFF
--- a/src/skillsaw/__main__.py
+++ b/src/skillsaw/__main__.py
@@ -530,7 +530,7 @@ def _run_fix(args):
 
     if args.model:
         config.llm.model = args.model
-    if args.max_iterations:
+    if args.max_iterations is not None:
         config.llm.max_iterations = args.max_iterations
 
     c = _ansi_colors()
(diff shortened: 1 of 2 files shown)


stbenjam:
@coderabbitai Review this please, what does "0" mean ? Is it unlimited iterations, or none?

PR #113, opened 2026-05-10
fix: handle null rules key in config

not-stbenjam:
## Summary
- Fix `AttributeError: 'NoneType' object has no attribute 'get'` crash when config YAML has `rules:` with no value (YAML null) or `rules: null`
- Same fix applied to `custom-rules` and `exclude` keys which had the identical vulnerability
- Changed `data.get("rules", {})` to `data.get("rules") or {}` so falsy values (None) are coalesced to empty dict/list

## Test plan
- [x] Added test for `rules: null` and bare `rules:` — verifies no crash and downstream methods work
- [x] Added test for `custom-rules: null` and `exclude: null` — verifies empty list fallback
- [x] All 466 existing tests pass
- [x] Lint clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration parsing to handle empty/null values for rules, custom-rules, exclude, and strict—applying sensible defaults (e.g., treating null/empty rules as empty sets and strict:null as false) and returning a clear error when the YAML top-level is not a mapping.
* **Tests**
  * Added coverage verifying robust handling of null/empty config entries and the non-mapping YAML error.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/113)

### DIFF
--- a/src/skillsaw/config.py
+++ b/src/skillsaw/config.py
@@ -41,9 +41,9 @@ class LinterConfig:
             raise ValueError(f"Failed to load config from {config_path}: {e}")
 
         return cls(
-            rules=data.get("rules", {}),
-            custom_rules=data.get("custom-rules", []),
-            exclude_patterns=data.get("exclude", []),
+            rules=data.get("rules") or {},
+            custom_rules=data.get("custom-rules") or [],
+            exclude_patterns=data.get("exclude") or [],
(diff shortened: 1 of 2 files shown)


not-stbenjam:
Closing as duplicate of #120 which provides more comprehensive null/wrong-type config validation.

PR #130, opened 2026-05-10
fix: correct line numbers when instructions entries are skipped

not-stbenjam:
## Summary

- Fix wrong line numbers reported by `_extract_instructions` when earlier `instructions:` YAML keys are skipped (empty, None, or non-string values)
- The root cause was using `len(results)` to find the nth `instructions:` key, but `len(results)` only counts validated entries -- skipped entries were not counted, causing `_find_nth_key_line` to resolve to the wrong YAML key
- Introduce an `instructions_keys_seen` counter that tracks ALL encountered `instructions:` keys (including skipped ones)

## Test plan

- [x] Added test: empty `reviews.instructions` followed by valid `path_instructions` entry -- line number must point to the path entry, not the empty one
- [x] Added test: null `reviews.instructions` followed by valid `path_instructions` entry
- [x] Added test: multiple skipped entries (empty `reviews.instructions` + empty path entry) with two valid path entries -- both line numbers must be correct
- [x] All 840 existing tests pass
- [x] `make lint` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of line number assignment for instruction configurations in YAML files. The system now correctly maps instruction entries when multiple configurations are present or when entries contain empty values, providing precise location references.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/130)

### DIFF
--- a/src/skillsaw/rules/builtin/content_analysis.py
+++ b/src/skillsaw/rules/builtin/content_analysis.py
@@ -396,18 +396,10 @@ def _extract_instructions(data: Any, raw: str) -> List[Tuple[str, str, Optional[
     if not isinstance(data, dict):
         return results
 
-    # Track the total number of `instructions:` keys encountered in the YAML
-    # (including skipped ones) so that _find_nth_key_line resolves to the
-    # correct occurrence.  Using len(results) would miscount when earlier
-    # entries are empty/None/non-string and therefore not added to results.
-    instructions_keys_seen = 0
-
     # reviews.instructions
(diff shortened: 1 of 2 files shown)


not-stbenjam:
Closing in favor of a centralized ruamel.yaml-based approach to YAML line number handling. The nth-key-line regex scanning is fragile — ruamel.yaml's round-trip parser will give us accurate line numbers for any key at any nesting level.

PR #136, opened 2026-05-10
fix: prevent line map collision for repeated YAML keys

not-stbenjam:
## Summary

- Fixed `_build_frontmatter_line_map` in the openclaw rule to use first-occurrence-wins semantics (`dict.setdefault`) instead of unconditional assignment
- When the same key name appears at different nesting levels (e.g. `os` at top-level and inside an `install` entry), the line map now records the first occurrence, preventing nested duplicates from silently overwriting the correct line number
- Added 3 tests: two unit tests for the line map function and one end-to-end test verifying violation line numbers are correct with duplicate keys

## Test plan

- [x] `test_line_map_first_occurrence_wins` -- verifies `os` key retains the first (top-level) line number
- [x] `test_line_map_first_occurrence_wins_bins` -- verifies `bins` under `requires` is not overwritten by `bins` under `install`
- [x] `test_top_level_os_line_with_nested_install_os` -- end-to-end test that a top-level `os` violation points to line 6 (not line 8)
- [x] All 467 existing + new tests pass
- [x] `make lint` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved manifest validation error reporting to correctly identify the line number of YAML keys when duplicates appear at different nesting levels. The system now accurately tracks the first occurrence of each key, ensuring precise error location guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/136)

### DIFF
--- a/src/skillsaw/rules/builtin/openclaw.py
+++ b/src/skillsaw/rules/builtin/openclaw.py
@@ -12,7 +12,15 @@ from skillsaw.rules.builtin.agentskills import _parse_skill_md
 
 
 def _build_frontmatter_line_map(skill_md: Path) -> Dict[str, int]:
-    """Map YAML key names to their line numbers in SKILL.md frontmatter."""
+    """Map YAML key names to their line numbers in SKILL.md frontmatter.
+
+    When the same key name appears at different nesting levels (e.g. ``os``
+    under the top-level openclaw block *and* inside an ``install`` entry),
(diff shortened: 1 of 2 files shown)


not-stbenjam:
Closing in favor of a centralized ruamel.yaml-based approach to YAML line number handling (see upcoming PR). The flat Dict[str, int] line map with setdefault was trading one wrong answer for another — the proper fix is nesting-aware parsing.

PR #151, opened 2026-05-13
feat: add per-rule excludes and inline suppression directives

not-stbenjam:
## Summary
- Add per-rule `excludes` config: glob patterns in `.skillsaw.yaml` to skip specific files for specific rules (e.g. `excludes: ["skills/legacy/**"]`)
- Add inline HTML comment directives: `<!-- skillsaw-disable rule-id -->` / `<!-- skillsaw-enable -->` and `<!-- skillsaw-disable-next-line rule-id -->` for surgical suppression in markdown
- Both mechanisms filter violations in the linter's dispatch logic after rules run, keeping rule implementations untouched
- Version bump to 0.8.3

## Test plan
- [x] Tests for per-rule excludes (matching patterns skip rule, non-matching still lint)
- [x] Tests for inline disable/enable pairs
- [x] Tests for disable-next-line
- [x] Tests for multiple rule IDs in one directive
- [x] Tests for unclosed disable (rest of file)
- [x] Tests for enable-all
- [x] `make test` passes (1113 tests)
- [x] `make lint` passes
- [x] Self-lint passes
- [x] Tested against openshift-eng/ai-helpers (exit 0)

## Release Notes

* **New Features**
  * Configure file exclusion patterns on a per-rule basis to skip specific rules for targeted files
  * Suppress violations inline in markdown using HTML comment directives: `skillsaw-disable`, `skillsaw-enable`, and `skillsaw-disable-next-line` for granular control

* **Documentation**
  * Updated configuration guide with per-rule excludes and inline suppression instructions

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/151)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,8 @@ Keep your skills sharp. A linter with built-in content intelligence for [agentsk
 - [Configuration](#configuration)
   - [Version Pinning](#version-pinning)
   - [Exclude Patterns](#exclude-patterns)
+  - [Per-Rule Excludes](#per-rule-excludes)
+  - [Inline Suppression](#inline-suppression)
   - [Content Paths](#content-paths)
 - [Builtin Rules](#builtin-rules)
 - [Autofixing](#autofixing)
@@ -366,6 +368,65 @@ exclude:
   - "generated/**"
   - "node_modules/**"
 ```
+
+### Per-Rule Excludes
+
+Exclude specific files from a single rule using the `excludes` key in the
+rule's config:
+
+```yaml
+rules:
+  content-weak-language:
+    enabled: true
+    excludes:
+      - "docs/legacy/**"
+      - "CHANGELOG.md"
+```
+
+This is useful when a rule produces false positives on specific files but
+you still want it enabled globally. Per-rule excludes use the same glob
+syntax as global `exclude` patterns.
+
+### Inline Suppression
+
+Suppress specific rules on specific lines using HTML comment directives
+directly in your markdown files:
+
+```markdown
+<!-- skillsaw-disable content-weak-language -->
+This section intentionally uses informal language.
+<!-- skillsaw-enable content-weak-language -->
+```
+
+Suppress a single line:
+
+```markdown
+<!-- skillsaw-disable-next-line content-tautological -->
+Follow best practices for error handling.
+```
+
+Suppress multiple rules at once:
+
+```markdown
+<!-- skillsaw-disable content-weak-language, content-tautological -->
+```
+
+Re-enable all suppressed rules:
+
+```markdown
+<!-- skillsaw-enable -->
+```
+
+Multi-line HTML comments are also supported:
+
+```markdown
+<!--
+    skillsaw-disable content-weak-language
+-->
+```
+
+Inline suppression only affects rules that are already enabled. It cannot
+be used to enable a normally disabled rule.
 
 ### Content Paths
 
--- a/src/skillsaw/suppression.py
+++ b/src/skillsaw/suppression.py
@@ -123,7 +123,7 @@ class SuppressionMap:
     # Maps file line number -> set of suppressed rule IDs
     _suppressed_lines: Dict[int, FrozenSet[str]] = field(default_factory=dict)
     # Set of lines where ALL rules are suppressed (empty rule list in disable)
-    _fully_suppressed_lines: Set[int] = field(default_factory=set)
+    _fully_suppressed_lines: FrozenSet[int] = field(default_factory=frozenset)
 
     def is_suppressed(self, rule_id: str, file_line: int) -> bool:
         """Check if a rule is suppressed at a given file line number."""
@@ -146,7 +146,6 @@ def build_suppression_map(content: str, line_offset: int = 0) -> SuppressionMap:
     Returns:
         SuppressionMap that can check if a rule is suppressed at a given line.
     """
-    total_lines = content.count("\n") + (1 if content and not content.endswith("\n") else 0)
     if content.endswith("\n"):
         total_lines = content.count("\n")
     else:
--- a/src/skillsaw/suppression.py
+++ b/src/skillsaw/suppression.py
@@ -20,11 +20,11 @@ from __future__ import annotations
 import re
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Dict, FrozenSet, List, Optional, Set, Tuple
+from typing import Dict, FrozenSet, List, Optional, Set
 
-# Matches <!-- skillsaw-disable rule-a, rule-b -->
+# Matches <!-- skillsaw-disable rule-a, rule-b --> or <!-- skillsaw-disable -->
 _DISABLE_RE = re.compile(
-    r"<!--\s*skillsaw-disable\s+([\w,\s-]+?)\s*-->",
+    r"<!--\s*skillsaw-disable\s*([\w,\s-]*?)\s*-->",
     re.IGNORECASE,
 )
 
@@ -77,10 +77,11 @@ def build_suppression_map(content: str, line_offset: int = 0) -> SuppressionMap:
         SuppressionMap that can check if a rule is suppressed at a given line.
     """
     lines = content.splitlines()
-    total_lines = len(lines)
 
     # Track currently disabled rule IDs
     disabled: Set[str] = set()
+    # Whether "disable all" is currently active (bare <!-- skillsaw-disable -->)
+    disable_all_active: bool = False
 
     # Per-line suppression data
     suppressed_lines: Dict[int, Set[str]] = {}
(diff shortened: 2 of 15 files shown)


src/skillsaw/suppression.py
```diff
@@ -0,0 +1,156 @@
...
+
+import re
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Dict, FrozenSet, List, Optional, Set, Tuple
+
+# Matches <!-- skillsaw-disable rule-a, rule-b -->
+_DISABLE_RE = re.compile(
+    r"<!--\s*skillsaw-disable\s+([\w,\s-]+?)\s*-->",
+    re.IGNORECASE,
```
stbenjam: You have handrolled your own HTML parsing, is there not a more reliable way to do this? Would you, for example detect multi-line comments?

```
<!--
    skillsable-enable
                       some-other-rule-->
```

not-stbenjam: Good catch -- the hand-rolled line-by-line regex approach would indeed miss multi-line comments like your example. I've rewritten the parser to use Python's `html.parser.HTMLParser` instead. The `HTMLParser.handle_comment()` callback receives the full comment text (even across multiple lines), and `getpos()` tracks the line number. The comment text is whitespace-normalized before matching directives, so formats like:

```html
<!--
    skillsaw-enable
                       some-other-rule-->
```

are now handled correctly. Added 6 new test cases covering multi-line disable/enable, disable-next-line, bare disable-all, multiple rules, and the exact whitespace pattern from your example.

stbenjam:
This whole PR needs docs changes to explain how to exclude in rules, it should be in README.md.

src/skillsaw/suppression.py
```diff
@@ -0,0 +1,264 @@
...
+
+    <!-- skillsaw-disable rule-id -->
+    ...suppressed content...
+    <!-- skillsaw-enable rule-id -->
+
+    <!-- skillsaw-disable-next-line rule-id -->
+    This single line is suppressed.
+
+    <!-- skillsaw-disable rule-a, rule-b -->
+    <!-- skillsaw-enable -->  (re-enables all)
```
stbenjam, line 15: Does the code allow swapping this, e.g. to enable a normally disabled rule? Does it make sense to do this? You do not have to implement this; it is a question whether it makes sense, or just global enable + scoped excludes  is sufficient

``` 
    <!-- skillsaw-enable rule-id -->
    ...suppressed content...
    <!-- skillsaw-disable -->
```


What about customizing a rule for a specific section, e.g. for THIS specific section, never allow it to grow beyond N tokens, etc.

stbenjam:
Do we have sufficient logging when in --verbose that these rules are being spotted invoked? It should be observable in verbose logging what's going on

src/skillsaw/suppression.py
not-stbenjam: Good questions. Current behavior:

**Can \`skillsaw-enable\` activate a normally disabled rule?**
No — inline suppression only works on rules that are already enabled via config. The directives are a _scoping_ mechanism (narrow where an enabled rule fires), not an _activation_ mechanism. This is documented in the README: "Inline suppression only affects rules that are already enabled. It cannot be used to enable a normally disabled rule."

I think this is the right default. Enabling a rule inline would mean the file silently changes its own lint behavior without any signal in `.skillsaw.yaml`, which makes CI results harder to reason about. Global enable + scoped excludes covers the practical cases: "I want this rule everywhere except here."

**Per-section rule customization (e.g. per-section token budgets)?**
This would require a richer directive syntax — something like `<!-- skillsaw-configure content-instruction-budget max_tokens=500 -->`. It's a compelling idea for power users but adds significant complexity to the parser and raises questions about config precedence. I'd suggest deferring this to a follow-up issue if there's concrete demand, and keeping this PR focused on the suppress/exclude primitives.

Happy to revisit either decision if you feel differently.

not-stbenjam:
Added README documentation for per-rule excludes and inline suppression in 6c7c640. The new sections cover:
- Per-rule `excludes` config with glob patterns
- `skillsaw-disable` / `skillsaw-enable` block suppression
- `skillsaw-disable-next-line` single-line suppression
- Multi-line HTML comment support
- Multiple rule suppression

not-stbenjam:
Added verbose logging for suppression and exclude activity in 6c7c640. When running with `--verbose`, the linter now logs each suppressed violation with the reason:

```
Suppressed content-weak-language      docs/legacy/guide.md (per-rule exclude)
Suppressed content-tautological       CLAUDE.md:15 (inline directive)
Suppressed plugin-structure           vendor/plugin.yaml (global exclude)
Filtered 3 of 12 violations via excludes/suppression
```

src/skillsaw/suppression.py
```diff
@@ -0,0 +1,156 @@
...
+    Args:
+        content: Full file content (including frontmatter etc.)
+        line_offset: Offset to add to body-relative line numbers to get file lines.
+                     For files without frontmatter this is 0.
+
+    Returns:
+        SuppressionMap that can check if a rule is suppressed at a given line.
+    """
+    lines = content.splitlines()
+    total_lines = len(lines)
```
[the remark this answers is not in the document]
not-stbenjam: Fixed in e2acba9 — `total_lines` is now used in the `range()` loop on line 183.

src/skillsaw/suppression.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # Track currently disabled rule IDs
+    disabled: Set[str] = set()
+
+    # Per-line suppression data
+    suppressed_lines: Dict[int, Set[str]] = {}
+    fully_suppressed_lines: Set[int] = set()
```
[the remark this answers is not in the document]
not-stbenjam: Fixed in 33f7703 — `fully_suppressed_lines` is now populated when `disable_all_active` is true (bare `skillsaw-disable` without rule IDs).

src/skillsaw/suppression.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # Lines where disable-next-line applies
+    next_line_rules: Optional[List[str]] = None
+
+    for line_num_0, line in enumerate(lines):
+        file_line = line_num_0 + 1  # 1-based
```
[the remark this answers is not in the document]
not-stbenjam: Fixed in 33f7703 — `line_offset` is now applied: `file_line = content_line + line_offset` (line 185).

src/skillsaw/suppression.py
not-stbenjam: This was addressed in 33f7703 — the parser now uses Python's `html.parser.HTMLParser` which correctly handles multi-line comments. Tests cover the exact whitespace pattern from your example.

README.md
```diff
@@ -367,6 +369,65 @@ exclude:
...
+```markdown
+<!-- skillsaw-disable content-weak-language -->
+This section intentionally uses informal language.
+<!-- skillsaw-enable content-weak-language -->
+```
+
+Suppress a single line:
+
+```markdown
+<!-- skillsaw-disable-next-line content-tautological -->
```
stbenjam, line 404: I am not sure optimizing this makes sense, we disable/enable guards are probably sufficient, what do you think?

README.md
```diff
@@ -367,6 +369,65 @@ exclude:
...
+
+### Inline Suppression
+
+Suppress specific rules on specific lines using HTML comment directives
+directly in your markdown files:
+
+```markdown
+<!-- skillsaw-disable content-weak-language -->
+This section intentionally uses informal language.
+<!-- skillsaw-enable content-weak-language -->
```
stbenjam, line 398: If someone does a disable, and never an enable - that only applies to that file's scope right? Do we test this at all?

not-stbenjam: Yes, correct — an unclosed `disable` suppresses the rest of that file only. Each file gets its own `SuppressionMap` built from scratch, so there's no cross-file bleed.

We have tests for this:
- `test_unclosed_disable_suppresses_rest_of_file` (unit, line 71) — verifies lines after the directive are suppressed and lines before are not
- `test_disable_all_unclosed` (unit, line 157) — same for bare `<!-- skillsaw-disable -->` without rule IDs
- `test_unclosed_disable_suppresses_rest_of_file` (integration, line 406) — end-to-end with the linter

README.md
not-stbenjam: Agreed — \`disable-next-line\` is a convenience but not essential. The disable/enable guards already cover the same use case, just more verbosely:

```markdown
<!-- skillsaw-disable-next-line content-tautological -->
Follow best practices for error handling.
```

vs.

```markdown
<!-- skillsaw-disable content-tautological -->
Follow best practices for error handling.
<!-- skillsaw-enable content-tautological -->
```

That said, \`disable-next-line\` is the standard pattern in eslint, rubocop, golangci-lint, etc., so users will instinctively reach for it. The implementation cost is minimal (6 lines in the parser). I'd lean toward keeping it for familiarity, but happy to remove it if you'd rather keep the surface area smaller.

PR #152, opened 2026-05-13
feat: fix false positives and add default exclude patterns

not-stbenjam:
## Summary
- Fix `content-negative-only` false positives on scope boundary patterns like "Don't use this when:" -- these are scope definitions, not behavioral prohibitions
- Reduce `content-weak-language` severity to INFO for `skill-ref` content blocks, since hedging language is appropriate in supplementary reference material
- Add default exclude patterns (`**/template/**`, `**/templates/**`, `**/_template/**`) to `LinterConfig.default()` so template scaffolding files are not linted by default
- Fix YAML serialization (`_yaml_value`) to properly quote glob patterns containing `*` and handle block-style list output without trailing whitespace

## Test plan
- [x] Tests for negative-only scope boundary skip (4 tests: colon, star, no-apostrophe variants, and positive case still flagged)
- [x] Tests for weak-language reference file severity reduction (2 tests: INFO for refs, WARNING for non-refs)
- [x] Tests for default exclude patterns in config (4 tests: patterns present, non-empty, user override, for_init inclusion)
- [x] `make test` passes (1104 passed, 14 skipped)
- [x] `make lint` passes
- [x] `make update` passes (regenerated docs and example config)

## Summary by CodeRabbit

* **New Features**
  * Default exclude patterns added to skip template-like directories during linting.
  * Content rules now classify reference-based content with lower-severity info findings and better recognize scope-bound negation statements.

* **Documentation**
  * Example config updated to v0.8.3 and README clarified: default excludes apply across rules and are fully replaced when users supply their own exclude list.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/152)

### DIFF
--- a/.skillsaw.yaml.example
+++ b/.skillsaw.yaml.example
@@ -1,7 +1,7 @@
 # skillsaw configuration
 # https://github.com/stbenjam/skillsaw
 
-version: "0.8.2"
+version: "0.8.3"
 
 rules:
 
@@ -229,7 +229,10 @@ rules:
 custom-rules: []
 
 # Exclude patterns (glob format)
-exclude: []
+exclude:
+    - "**/template/**"
+    - "**/templates/**"
+    - "**/_template/**"
 
 # Additional markdown files to run content rules on (glob format)
 content-paths: []
--- a/.skillsaw.yaml.example
+++ b/.skillsaw.yaml.example
@@ -269,8 +269,11 @@ rules:
 custom-rules: []
 
 # Exclude patterns (glob format)
-# Default: **/template/**, **/templates/**, **/_template/**
-exclude: []
+# Use exclude: [] to disable all excludes including defaults
+exclude:
+    # - "**/template/**"
+    # - "**/templates/**"
+    # - "**/_template/**"
 
 # Additional markdown files to run content rules on (glob format)
 content-paths: []
--- a/.skillsaw.yaml.example
+++ b/.skillsaw.yaml.example
@@ -83,6 +83,7 @@ rules:
   mcp-prohibited:
     enabled: false
     severity: error
+    # allowlist: []
 
   # .claude/rules/ files must be markdown with valid optional paths frontmatter
   rules-valid:
@@ -108,6 +109,11 @@ rules:
   agentskill-structure:
     enabled: false
     severity: warning
+    # allowed_dirs:
+    #   - assets
+    #   - evals
+    #   - references
+    #   - scripts
 
   # Require evals/evals.json for each skill (opt-in)
   agentskill-evals-required:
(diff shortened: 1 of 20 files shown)


stbenjam:
Do excludes work on custom rules? Do you have integration tests for this behavior? Have you updated the README.md with docs?

not-stbenjam:
Yes, excludes work on all rules including custom rules -- the exclude filtering happens at the linter level in `Linter._is_excluded()` after all rule results are collected, so it applies uniformly.

In 60226ea:
- Added an integration test (`test_custom_rule_respects_exclude_patterns`) that creates a custom rule reporting violations for every markdown file, then verifies that files in excluded directories are filtered out.
- Updated the README Exclude Patterns section to document that default patterns include template directories and that excludes apply to all rules including custom rules.

src/skillsaw/rules/builtin/content_rules.py
```diff
@@ -60,16 +60,21 @@ def llm_fix_prompt(self):
             "- Preserve markdown formatting"
         )
 
+    # Categories where hedging language is acceptable (supplementary material)
+    _REFERENCE_CATEGORIES = frozenset({"skill-ref"})
```
stbenjam: Hm why isn't skill-ref an enum type? Does it exist? Why are these bare strings?

.skillsaw.yaml.example
```diff
@@ -269,7 +269,10 @@ rules:
 custom-rules: []
 
 # Exclude patterns (glob format)
-exclude: []
+exclude:
+    - "**/template/**"
+    - "**/templates/**"
+    - "**/_template/**"
```
stbenjam: Why aren't these the defaults, and have the config generator put them here as commented out for consistency?  That way if they omit exclude from `.skillsaw.yaml.example` or the file entirely the defaults will apply

src/skillsaw/rules/builtin/content_rules.py
not-stbenjam: Good catch — replaced the bare string check with `isinstance(cf, SkillRefBlock)` so it uses the typed block class instead of matching on a magic string.

.skillsaw.yaml.example
not-stbenjam: Fixed — defaults now apply when the key is omitted (or null). Users can use `exclude: []` for no excludes. The example config shows the defaults as a comment for discoverability.

.skillsaw.yaml.example
```diff
@@ -269,6 +269,7 @@ rules:
 custom-rules: []
 
 # Exclude patterns (glob format)
+# Default: **/template/**, **/templates/**, **/_template/**
 exclude: []
```
stbenjam: For rules, we put the default values as a comment.

```
exclude:
#  - "**/template/**
```

And with this format, how would a user override it to mean "don't exclude anything, not even templates"

not-stbenjam: Updated — defaults now show as commented-out items under `exclude:`, matching the rule format. To disable all excludes including defaults, use `exclude: []`.

stbenjam: Is there a test for the `exclude: []` format?

not-stbenjam: Added `test_empty_exclude_disables_all_defaults` — verifies that `exclude: []` results in an empty pattern list with no defaults applied.

stbenjam:
Why is make update removing the commented fields? They should stay but be generated

PR #188, opened 2026-05-14
Optimize lint performance for large repos

not-stbenjam:
## Summary

- Cache `_strip_fenced_code_blocks` result in `read_body()` — was being called 2851 times redundantly across rules
- Combine 27 secret-detection regex patterns into a single pre-filter regex, reducing per-line regex calls from 27 to 1 for non-matching lines (the common case)
- Same combined-regex optimization for banned-references rule
- Pre-compile all detector patterns (hedging, vagueness, tautological, contradiction) that were previously compiled on each call via `re.finditer(str_pattern, ...)`
- Cache `resolved_path` on ContentBlock to avoid repeated `Path.resolve()` syscalls (was 9008 calls)
- Single-pass the actionability score loop instead of 3 separate `sum()` passes

**Benchmark on openshift-eng/ai-helpers** (259 content blocks, 80K total lines, 34 rules):
- Before: **2.17s** (3.8M regex searches)
- After: **1.95s** (833K regex searches — 78% reduction)

## Test plan

- [x] All 1418 tests pass (`make test`)
- [x] `make lint` passes
- [x] `make update` — no unexpected changes
- [x] `skillsaw lint` against openshift-eng/ai-helpers — exit 0, identical results
- [x] Added 5 new tests in `TestReadBodyCache` to verify caching correctness and invalidation

## Summary by CodeRabbit

* **Refactor**
  * Improved content analysis with precompiled patterns and short-circuit checks for faster, more reliable detections.
  * Added caching for content blocks (resolved paths and stripped-body results) with proper invalidation on updates to reduce repeated work.

* **Tests**
  * Added tests to validate read-body caching, cache invalidation on write, and cached resolved-path consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stbenjam/skillsaw/pull/188)

### DIFF
--- a/src/skillsaw/rules/builtin/content_analysis.py
+++ b/src/skillsaw/rules/builtin/content_analysis.py
@@ -111,57 +111,144 @@ class InstructionBudget:
 # --- Patterns ---
 
 _HEDGING = [
-    (r"\btry to\b", "Remove 'try to' — state the action directly"),
+    (re.compile(r"\btry to\b", re.IGNORECASE), "Remove 'try to' — state the action directly"),
     (
-        r"\bconsider\s+(?:using|adding|implementing|creating|moving|switching|enabling)\b",
+        re.compile(
+            r"\bconsider\s+(?:using|adding|implementing|creating|moving|switching|enabling)\b",
+            re.IGNORECASE,
+        ),
         "Replace 'consider X' with 'do X' or remove",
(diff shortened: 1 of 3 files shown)


stbenjam:
~10% not really worth it.  I think the loading time I'm seeing is from the docker image

PR #215, opened 2026-05-27
Reproducer: .mdc line offset is wrong (comments on wrong line)

stbenjam:
## Reproducer for #213

Weak language on **file line 15** of `.cursor/rules/python.mdc`, but skillsaw reports it as **line 10** (off by 5 — the frontmatter block size).

The Lint Review action should post inline comments on line 10, visibly demonstrating the bug.

**DO NOT MERGE** — close after observing.


stbenjam:
Fixed!

PR #292, opened 2026-06-11
Settings/hooks/MCP JSON blocks are config, not content

stbenjam:
## Summary

`skillsaw lint` flagged `.claude/settings.local.json` with `content-cognitive-chunks` ("No headings in instruction file — add section headings for cognitive chunking") — a JSON settings file linted as prose.

**Root cause**: `HooksBlock`, `McpBlock`, and `SettingsBlock` inherited `FileContentBlock` (a `ContentBlock`), and `content_blocks()` — the source for all 17 content-quality rules plus `context-budget` — excluded hooks and MCP via an isinstance blocklist that settings was never added to.

**Fix**: encode the prose-vs-config distinction in the type hierarchy instead of a blocklist. A new `JsonConfigBlock(LintTarget)` base holds the JSON parse machinery (`parse_error`/`raw_data`, previously copy-pasted identically in all three classes), token estimation for `skillsaw tree`, and the tree label. The three config blocks subclass it, and `content_blocks()` becomes a plain `find(ContentBlock)` with no exclusion list to forget to update.

**Class-of-bug audit**: all remaining `ContentBlock` subclasses are genuinely prose (instruction markdown, CodeRabbit YAML instruction fragments, promptfoo prompts, skill references) — settings/hooks/MCP were the only structured config files in the prose hierarchy. Dedicated rules (`settings-dangerous`, `hooks-*`, `mcp-*`) still reach the config blocks via `find(SettingsBlock)` etc.

**Why CI never caught it**: `settings.local.json` is gitignored, so it only appears in local working trees; the repo's own fixture settings files were all below the rule's 100-character threshold.

Also records the `ContentBlock` vs `JsonConfigBlock` contract in `.apm/instructions` (regenerated via `make update`).

## Test plan

- New regression fixture `tests/fixtures/regression/settings-not-content/` with realistic, threshold-exceeding `settings.json` + `settings.local.json`
- `test_settings_files_skip_content_rules` — fails on the old code, passes with the fix
- `test_settings_files_still_get_settings_rules` — guards the other direction: dangerous settings content is still flagged by the settings rules
- `make test` — 1985 passed
- `make lint`, `make update` clean
- `skillsaw lint` against openshift-eng/ai-helpers — exit 0, and the reported `settings.local.json` warning is gone
- `skillsaw tree` output for settings/hooks/MCP files unchanged (labels and token counts)


not-stbenjam:
@coderabbitai plz review

PR #300, opened 2026-06-13
Fix confirmed bugs from the architecture/quality review

not-stbenjam:
A batch of confirmed bugs surfaced by a multi-angle architecture/quality review of the codebase, each fixed with a regression test. **+37 tests, full suite green (2119 passed), `make lint` clean, self-lint and `openshift-eng/ai-helpers` both exit 0.**

The companion review report is filed as a tracking issue (the "Fable sweep"); this PR implements the confirmed-bug subset. Large architectural refactors from the report (block-hierarchy extraction, `__main__.py` split, LLM-orchestrator carve-out) are intentionally **not** bundled here — they belong in their own PRs.

## Suppression
- **Directives inside markdown code fences no longer fire.** A `# skillsaw-disable` in a fenced example previously suppressed every rule for the rest of the file. `#` directives are now skipped inside fenced/indented code blocks (headings stay honored).
- **Word boundary + start anchor on directives.** Prose like `<!-- TODO: document skillsaw-disable -->` and `skillsaw-disabledthing` are now inert instead of disabling all rules.

## Config
- **Friendly errors instead of tracebacks** for malformed configs: non-mapping root (`- a`), `llm: gpt-4`, `content-paths: "x"`, mistyped `llm`/`exclude` items — all now raise `ValueError` and exit 1 cleanly.
- **Warnings** for a missing `version` field (silently disables newer rules) and unknown top-level keys.
- `encoding="utf-8"` on config reads; default-rules mapping cached (was rebuilt per-violation on the filter hot path); `find_config` now checks the filesystem root.

## Baseline
- **Baseline in a parent directory now matches when linting a subdirectory** — fingerprints are computed relative to the directory that owns the baseline file (previously nothing matched and every entry went stale).
- **Ratchet fingerprint collisions fixed** via an optional `metric` discriminator (context-budget emitted whole-file + per-description token violations with identical fingerprints). Empty metric preserves the legacy fingerprint, so existing baselines keep matching.
- `find_baseline` now checks the filesystem root.

## CodeRabbit / promptfoo
- **`CodeRabbitContentBlock.write_body` no longer crashes** on a `path` glob containing a dot (`**/*.py` — the common case). Uses index-based, bracket-aware yaml paths; a bad path degrades to a no-op.
- **Correct line numbers** for path_instructions/tools/custom_checks (was an nth-occurrence heuristic that mis-attributed lines when other `instructions` keys preceded them).
- **promptfoo prompt line numbers** distinguish plain vs block scalars (plain one-line prompts were one line too low).

## CLI / linter / formatters
- Missing/broken `custom-rules` files raise a friendly error instead of a traceback; each custom rule loads under a **unique module name**; custom rules now honor their `since` version gate.
- **`--skip-rule` validates IDs** (a typo previously left the rule running).
- `--output FORMAT:` with an empty path errors up front; duplicate `--output` targets with conflicting formats error instead of silently dropping one.
- **SARIF URIs** use forward slashes and omit `uriBaseId` for paths outside the repo root.
- `extract_section` matches CRLF content.

## Summary by CodeRabbit

* **New Features**
  * Emits non-fatal configuration warnings at runtime.
  * Distinguishes multiple ratchet violations per file via optional metrics (when emitted).
* **Bug Fixes**
  * Improves baseline fingerprinting/filtering consistency (including stability when baselines/configs are nested) and avoids metric collisions.
  * Tightens configuration validation and discovery, including “Version Pinning” behavior.
  * Validates `--output FORMAT:PATH` (errors on missing paths) and improves SARIF URIs.
  * Fixes `skillsaw-disable*` directive handling in markdown code vs prose.
  * Hardens custom-rule loading and rejects unknown skip IDs; `--type` overrides now affect discovery.
* **Documentation**
  * Added “Version Pinning” guidance and updated CLI `--type` options (includes `apm`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,10 @@ unless you bump the version or explicitly enable them. Repos **without** a
 `.skillsaw.yaml` run all rules at the latest version — you get new rules
 automatically but may occasionally fail after a skillsaw upgrade.
 
+A hand-written config that omits `version` defaults to an early version (so
+newer rules are skipped) and prints a warning on each run — set `version`
+explicitly to silence it and control which rules apply.
+
 ### Exclude Patterns
 
 Skip files and directories using glob patterns:
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -18,7 +18,7 @@ Lint agent skills, plugins, and AI coding assistant context
 | `--strict` | Treat warnings as errors (exit with error code if warnings exist) |  |
 | `--format` | Output format for stdout (default: text) (choices: text, json, sarif, html, code-climate, gitlab) | `text` |
 | `--output` | Write output to FILE. Format is inferred from extension (.htm, .html, .json, .sarif, .txt) or set explicitly with a FORMAT: prefix (e.g. gitlab:report.json). Use the prefix when an extension is ambiguous (e.g. .json could be json or gitlab/code-climate). Can be specified multiple times. |  |
-| `--type` | Override auto-detected repository type (repeatable). Values: single-plugin, marketplace, agentskills, dot-claude, coderabbit. |  |
+| `--type` | Override auto-detected repository type (repeatable). Values: single-plugin, marketplace, agentskills, dot-claude, coderabbit, apm, promptfoo. |  |
 | `--rule` | Only run these rules (repeatable). Config still comes from .skillsaw.yaml. |  |
 | `--skip-rule` | Skip these rules (repeatable). Cannot be combined with --rule. |  |
 | `--no-baseline` | Ignore baseline file even if .skillsaw-baseline.json exists |  |
--- a/src/skillsaw/__main__.py
+++ b/src/skillsaw/__main__.py
@@ -46,6 +46,12 @@ def _get_version() -> str:
         return __version__
 
 
+def _emit_config_warnings(config) -> None:
+    """Print non-fatal config-load warnings (missing version, unknown keys)."""
+    for warning in getattr(config, "warnings", []):
+        print(f"Warning: {warning}", file=sys.stderr)
+
+
 def _build_parser():
     """Build the main argument parser with all subcommands.
 
@@ -501,6 +507,7 @@ def _run_tree(args):
         except ValueError as e:
             print(f"Error loading config: {e}", file=sys.stderr)
             sys.exit(1)
+        _emit_config_warnings(config)
     else:
         config = LinterConfig.default()
 
@@ -668,6 +675,13 @@ def _run_lint(args):
         except ValueError as e:
             print(f"Error: {e}", file=sys.stderr)
             sys.exit(1)
+        if filepath in output_formats and output_formats[filepath] != fmt:
+            print(
+                f"Error: --output targets '{filepath}' with conflicting formats "
+                f"'{output_formats[filepath]}' and '{fmt}'",
+                file=sys.stderr,
+            )
+            sys.exit(1)
         output_formats[filepath] = fmt
 
     _handle_apply_patch_for_lint(args)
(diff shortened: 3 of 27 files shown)


not-stbenjam:
Tracking issue with the full review sweep (including the deferred architectural refactors and lower-severity items): #301

not-stbenjam:
Thanks @coderabbitai and @gemini-code-assist — all feedback addressed in 1d079fd:

| Finding | Resolution |
|---|---|
| **config.py** — `version:` (null) gates as 0.0.0 | `raw_version is None` now treated as missing: warns + defaults to `_DEFAULT_VERSION`. Test: `test_null_version_treated_as_missing`. |
| **config.py** — `custom-rules: [42]` reaches linter as raw `TypeError` | Validate items are strings in `from_file`. Test: `test_custom_rules_non_string_items_raises`. |
| **config.py** — cached `_default_rules()` mutably shared | `get_rule_config` now `copy.deepcopy`s the per-rule defaults. |
| **__main__.py** — `--output` dedup on raw string | Keyed on `Path(filepath).resolve()` so `report.json` / `./report.json` collide. |
| **suppression.py** — patterns not end-anchored (`skillsaw-disable.`) | Patterns end-anchored (`\s*$`) and switched to `.match`; optional rule group. Tests: `test_trailing_punctuation_not_a_directive`, `test_next_line_with_trailing_colon_not_a_directive`. |
| **content_analysis.py** — block-scalar headers `\|2-` / `>2+` missed | Regex handles indentation+chomping indicators in either order. |
| **test_baseline.py** — subdirectory assertion was happy-path only | Added failing-control with the wrong root. |
| **test_config.py** — `match=` dotted patterns unescaped (RUF043) | Escaped to raw strings. |
| **test_config.py** — warnings-equality test used identical inputs | Now differs only by a warning (unknown key), asserts `a == b`. |
| **test_custom_rules.py** — module-isolation test didn't prove isolation | Linter now registers each rule module in `sys.modules` under its unique name (rolling back on failure); test asserts distinct `sys.modules` entries. |

Full suite green (2123 passed), `make lint` clean.

stbenjam:
@codex review

stbenjam:
@codex are you there?

stbenjam:
@codex review

not-stbenjam:
Thanks @chatgpt-codex-connector — all four addressed in ddfc621:

| Finding | Resolution |
|---|---|
| **baseline.py** — adding `metric` to context-budget breaks pre-upgrade baselines | The whole-file budget violation now keeps its legacy (metric-less) fingerprint; only the per-description violation carries a metric to break the collision. Existing baselines keep matching. Test: `test_whole_file_keeps_legacy_fingerprint`. |
| **sarif.py** — root-relative paths lose `%SRCROOT%` | A non-absolute `file_path` is treated as already root-relative (`under_root=True`), so it keeps `uriBaseId` and a forward-slash URI. Test: `test_sarif_relative_path_is_root_relative`. |
| **config.py** — `[]`/`false`/`0` coerced to `{}` and accepted | Only `None` (empty document) becomes an empty config; other non-mappings reach the type check and raise. Tests: `test_falsy_non_mapping_roots_raise`, `test_empty_file_is_empty_config`. |
| **content_analysis.py** — block scalar with trailing comment (`- \| # system prompt`) misdetected | Block-scalar regex allows an optional trailing YAML comment. Test: `test_block_scalar_with_trailing_comment_line_is_correct`. |

Full suite green (2128 passed), `make lint` clean.

not-stbenjam:
Also applied the CodeRabbit nitpick in 5dd3d53 — chained the config-load exception (`raise ValueError(...) from e`) so the original YAML/IO/decode error is preserved as `__cause__`.

PR #304, opened 2026-06-21
Add streamable-http as a valid MCP transport type

nyechiel:
## Summary

- Adds `streamable-http` to the list of valid MCP transport types in the `mcp-valid-json` rule
- Like `http` and `sse`, `streamable-http` requires a `url` field
- Adds test fixtures and test cases for valid and missing-url scenarios

## Context

The [MCP specification](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http) defines `streamable-http` as a transport type, but skillsaw only recognized `stdio`, `http`, and `sse`. This caused false positive `mcp-valid-json` errors on repositories using `streamable-http` servers (e.g., remote MCP endpoints like Red Hat's Dataverse).

## Test plan

- [x] Existing tests pass (2084 passed)
- [x] New test: valid `streamable-http` config passes validation
- [x] New test: `streamable-http` missing `url` field is detected
- [x] `make update` passes (format, lint, self-lint)

## Summary by CodeRabbit

* **New Features**
  * Expanded MCP server type support to include `streamable-http`, which requires the `url` field in configuration to ensure valid server setup.

* **Tests**
  * Added test coverage for the new `streamable-http` server type, verifying both valid configurations and validation of required fields.

### DIFF
--- a/src/skillsaw/rules/builtin/mcp/valid_json.py
+++ b/src/skillsaw/rules/builtin/mcp/valid_json.py
@@ -15,8 +15,13 @@ from skillsaw.rules.builtin.utils import read_json
 class McpValidJsonRule(Rule):
     """Check that MCP configuration is valid JSON with proper structure"""
 
-    VALID_MCP_TYPES = ("stdio", "http", "sse")
-    REQUIRED_FIELDS_BY_TYPE = {"stdio": "command", "http": "url", "sse": "url"}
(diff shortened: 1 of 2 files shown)


stbenjam:
Thanks so much! I released 0.14.1 with this and some other bug fixes

PR #306, opened 2026-06-21
Add pluribus.md context file detection

caioribeiroclw-pixel:
## Summary
- detect root-level `pluribus.md` as an instruction/context format
- include it in the lint tree as a typed `PluribusMdBlock` with `pluribus-md` category
- add coverage for format detection and lint-tree block discovery

## Why
skillsaw already audits cross-agent context surfaces like AGENTS.md, CLAUDE.md, GEMINI.md, Cursor rules, Copilot instructions, and APM context files. `pluribus.md` is another shared context file used to sync project context across AI coding tools, so treating it as lintable instruction content keeps it visible instead of an untyped markdown file.

## Validation
- `python3 -m compileall -q src tests`
- `git diff --check`

I could not run pytest in this container because Python has neither pytest nor pip installed (`python3 -m pytest` -> `No module named pytest`; `python3 -m pip --version` -> `No module named pip`).

## Summary by CodeRabbit

* **New Features**
  * Added support for Pluribus repositories as an additional instruction format source.
  * The system now automatically detects and processes Pluribus instruction files at the repository root.

### DIFF
--- a/src/skillsaw/context.py
+++ b/src/skillsaw/context.py
@@ -37,6 +37,7 @@ HAS_GEMINI = "HAS_GEMINI"
 HAS_AGENTS_MD = "HAS_AGENTS_MD"
 HAS_KIRO = "HAS_KIRO"
 HAS_CLAUDE_MD = "HAS_CLAUDE_MD"
+HAS_PLURIBUS = "HAS_PLURIBUS"
 HAS_CODERABBIT = "HAS_CODERABBIT"
 ALL_INSTRUCTION_FORMATS = frozenset(
     {
@@ -46,6 +47,7 @@ ALL_INSTRUCTION_FORMATS = frozenset(
         HAS_AGENTS_MD,
         HAS_KIRO,
         HAS_CLAUDE_MD,
+        HAS_PLURIBUS,
         HAS_CODERABBIT,
     }
 )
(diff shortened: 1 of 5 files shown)


stbenjam:
Hey, thanks so much for this — really appreciate the contribution! 🙏

Quick question first: is `pluribus.md` from [this project](https://github.com/caioribeiroclw-pixel/pluribus)? Want to make sure I understand the format.

Good news is you can lint these today without any code change: just point skillsaw at them via [content paths](https://skillsaw.org/configuration/#content-paths) and they'll pick up all the content-quality rules. If there's a specific schema or convention to enforce, custom rules can handle that too. 

For formats built in by default, I try to stick to ones with broad adoption just to keep things focused — but I'll admit I might be underestimating how widely `pluribus.md` is used, can you share more about its adoption?

caioribeiroclw-pixel:
Thanks for the quick read — yes, it is from that Pluribus project.

And to be direct: adoption is still early. I should not imply `pluribus.md` is in the same bucket as `CLAUDE.md`, `AGENTS.md`, Cursor rules, etc. The reason I opened this was that Pluribus uses `pluribus.md` as a source-of-truth context file and then syncs/generated tool-specific files, so it behaves like an instruction/context surface that benefits from the same content-quality linting.

Given your built-in-format bar, I think your point is fair. If built-ins are reserved for broadly adopted conventions, this probably belongs in user config via `content_paths` / custom rules for now rather than as a default format.

Happy to close this PR, or convert the idea into a small docs/config example only if you think that would be useful. I don't want to add niche defaults to Skillsaw before there is enough external adoption.

caioribeiroclw-pixel:
Closing this myself after thinking about your adoption bar.

You're right that `pluribus.md` should not be a Skillsaw built-in before there is broad independent usage. I moved the integration path to Pluribus docs instead: teams can lint it through `content-paths` without asking Skillsaw to bless a niche default.

Docs added here: https://github.com/caioribeiroclw-pixel/pluribus/blob/main/docs/lint-pluribus-context-with-skillsaw.md

Thanks again for the quick correction — it was useful product signal.

PR #345, opened 2026-07-02
Add opt-in agentskill-description-length rule

not-stbenjam:
## What

Adds a new builtin rule, `agentskill-description-length`: an opinionated, configurable soft budget on SKILL.md `description` length. When enabled, it reports a **warning** on the `description` key's line whenever the parsed description exceeds `max_length` characters (default **256**), including the actual length and the threshold in the message. Missing, empty, or non-string descriptions are ignored — those remain `agentskill-valid` / `agentskill-description`'s job. No autofix: a description can't be mechanically shortened.

## Why a separate rule from spec validation

The agentskills.io spec's hard 1024-character limit is spec parity and stays untouched (enforced by the existing `agentskill-description` warning; `agentskill-valid` covers required fields/types). This rule adds *quality guidance* on top: skill descriptions are permanent context, loaded into every prompt for routing, so every character is paid on every request — and some ecosystems rank/route on only a description prefix, with truncation of long descriptions claimed for some routers. Keeping spec parity and opinionated guidance in separate rules means users who disagree with the soft budget can tune or disable it without losing spec validation. At default settings there is no overlapping output with the sibling rules (verified by an integration test).

## Why opt-in

`default_enabled = False` — this is a new opinionated rule and must not fire on existing codebases until users enable it (per the project rule to never force-enable a rule that could break existing users).

## Config

```yaml
rules:
  agentskill-description-length:
    enabled: true
    max_length: 256   # soft budget in characters
```

`repo_types`: agentskills, single-plugin, marketplace, dot-claude (same as the sibling agentskill rules).

## Test coverage

- **Unit tests** (`tests/test_agentskill_rules.py`, 7 new): under budget passes; over default budget warns with length + threshold + key line; exactly at `max_length` passes (boundary); configured `max_length` moves the firing point; folded `description: >` measured on the parsed string with the line pointing at the key; missing/non-string descriptions ignored; disabled-by-default in `LinterConfig.default()`.
- **Integration tests** (`tests/test_integration.py`, new `TestDescriptionLengthConfig` class + static fixture `tests/fixtures/config/description-length/` with four realistic skills at 343 / 136 / exactly 256 / folded-303 chars): silent by default; fires when enabled with correct message and line; boundary; `max_length: 100` override fires on a 136-char description that default does not; folded multiline; no duplicate reporting alongside `agentskill-valid` / `agentskill-description` for the same file.
- Added to `OPT_IN_RULES` and the `config/opt-in-rules` fixture so the every-rule-fires-somewhere coverage guard covers it.
- Full suite: **2147 passed**. `make lint` and `make update` clean (README, `.skillsaw.yaml.example`, and site docs regenerated; rule docs page added for `skillsaw explain`).

## ai-helpers verification

Cloned `openshift-eng/ai-helpers` and ran the branch build: **0 errors**, and `agentskill-description-length` does not fire (opt-in). Exit code is 0 with `--no-plugins` (matching a stock `pip install skillsaw`); without that flag the dev-venv-only `skillsaw-typos` plugin flags 27 pre-existing typos in ai-helpers content under their `strict: true` config — unrelated to this change and identical on main.

## Notes

- A previously interrupted attempt at this rule left no branch or PR behind, so the natural branch name `agentskill-description-length` was used.
- Config key is `max_length` (underscore), matching existing underscore-style keys like `allowed_dirs`.

## Summary by CodeRabbit

* **New Features**
  * Added a new optional rule to keep skill descriptions concise, with a default soft limit of 256 characters.
  * Added documentation and examples showing how to shorten descriptions or adjust the limit.

* **Bug Fixes**
  * Improved validation so overlong descriptions are reported clearly and measured from the parsed text.
  * Reduced duplicate reporting by ensuring this check appears once per issue.


stbenjam:
A new rule doesn't make sense, just make agentskill-description configurable length.  The docs can reccomend decreasing it for tight control around ~256, and strongly discourage more than 1024.

not-stbenjam:
Closing: direction changed — instead of a new rule, the existing agentskill-description rule will gain a configurable max_length (default 1024, unchanged behavior). Replacement PR to follow.

Issue #377, opened 2026-07-05
Bug: `--fail-on error` exits 0 even when errors are present


stbenjam:
## Description

`skillsaw lint --fail-on error` is supposed to make the process exit **non-zero** when there are error-level violations, so CI fails on bad files. In my pipeline it exits `0` even though skillsaw clearly printed error-level violations — so CI stays green and the bad file slips through.

## Steps to reproduce

1. A `CLAUDE.md` that triggers at least one **error**-severity rule.
2. Run:
   ```
   skillsaw lint --fail-on error .
   echo "exit: $?"
   ```
3. Observed: `exit: 0` (despite errors in the output).

## Expected

Exit code should be `1` (non-zero) whenever error-level violations are present, so `--fail-on error` can actually gate CI.

## Environment

- skillsaw: latest `main`
- OS: Ubuntu 22.04 (GitHub Actions runner)

This makes the `--fail-on` flag useless for CI gating — pretty important for us.

PR #386, opened 2026-07-05
Triage: allow PyPI egress so the agent can actually run skillsaw

stbenjam:
## What

The triage skill's bug-evaluation step instructs the agent to **reproduce by running skillsaw** (`.venv/bin/skillsaw lint …`). But the network allowlist was `defaults` + `openrouter.ai` — no PyPI — so the agent can't `pip install -e .` to get a runnable skillsaw.

The risk (visible in [#377](https://github.com/stbenjam/skillsaw/issues/377)): the triage presents a `$ skillsaw lint …` block with command output and cites `line 246`, but if it couldn't actually run skillsaw, that "reproduction" is **inferred from reading the code and formatted to look like a real run** — the exact "never assert what you didn't verify" failure. The conclusion was right this time; we don't want to rely on luck.

Adds the `python` network ecosystem (expands to `pypi.org` + `files.pythonhosted.org`) so reproduction is genuine.

```yaml
network:
  allowed:
    - defaults
    - openrouter.ai
    - python        # PyPI
```

## Note

The maintainer-only `triage-for-agent` label gate remains the primary control against a crafted `.skillsaw.yaml` (`custom-rules` = arbitrary code execution) — a human approves each issue before the agent runs. A follow-up could add `--no-custom-rules` to the reproduce step for defense-in-depth.

## Verification
- `gh aw compile` clean; lock allowlist now includes `pypi.org` + `files.pythonhosted.org`. `make verify-apm` passes.

---

<a href="https://app.devin.ai/review/stbenjam/skillsaw/pull/386" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


stbenjam:
Addressed @devin-ai-integration's review in \`a877b18\`:

- **Stale egress header** — fixed. The security-model comment now lists all three egress categories (GitHub/Copilot API, OpenRouter BYOK, and PyPI) instead of "GitHub only."
- **\`python\` preset breadth (conda domains)** — keeping the \`python\` ecosystem token deliberately. It's the identifier gh-aw's own compiler recommends over hand-listing \`pypi.org\`/\`files.pythonhosted.org\`, and the maintainability is worth the extra (unused) conda domains. Noting it as a conscious tradeoff per REVIEW.md rather than an oversight.

PR #392, opened 2026-07-06
WIP: PostToolUse plugin hook for real-time linting

stbenjam:
> **WIP** — not ready to merge. Opening for early review of the approach.

## What

Ships a Claude Code **PostToolUse hook** in the skillsaw plugin that lints the file just edited and feeds violations back to the agent in real time, using the repository's own `.skillsaw.yaml` (its `fail-on` bar and baseline). The hook is a thin wrapper around a new subcommand:

```
skillsaw hook post-tool-use   # reads the tool-call payload as JSON on stdin
```

Where the pre-commit hook gates a finished commit, this is live feedback on a single edit.

## Design decisions

- **No filename allowlist in the hook.** An early version kept its own list of "files skillsaw lints" and it immediately drifted (dropped `*.instructions.md`). The hook now just lints the edited file's repository and scopes the result to that file — skillsaw's discovery is the single source of truth. Files it doesn't recognize yield no violations and the hook stays silent.
- **The `hooks.json` matcher (`Edit|Write|MultiEdit|Update`) is the only gate** on which tools fire; the handler keys off `tool_input.file_path`, not tool name, so the two can't drift.
- **Fail-safe:** malformed input, a missing file, a broken config, or an internal error all exit `0` and never disrupt editing. Exits `2` (surfacing stderr to the agent) only for violations at/above the repo's `fail-on`.

## Dogfooding

`make update` propagates the hook into skillsaw's own `.claude/`, `.cursor/`, `.codex/`, and `.github/` agent contexts, and `.skillsaw.yaml` now sets `fail-on: warning` so prose warnings surface live while working on skillsaw itself.

## Tests

`tests/test_hook.py` (17 tests): violations surfaced/scoped, fail-on gating honored from repo config, instruction-file regression, and the fail-safe no-op paths.

## Open questions / WIP

- Per-edit context build happens for any file the matcher passes — skillsaw's normal single-file cost, but worth watching on large repo roots.
- No version bump yet (holding until this is out of WIP).
- Docs: added `docs/claude-code-hook.md`; README intentionally left slim.


stbenjam:
I think pre-commit hook is probably better for this.  The Claude hooks are neat, but requires skillsaw being in path, etc.

PR #399, opened 2026-07-06
Add content-instruction-drift: detect drifted near-duplicate sections across instruction files

not-stbenjam:
## What

A new content-intelligence rule, `content-instruction-drift`, that catches a failure mode every multi-assistant repo hits: someone copies a section between CLAUDE.md, AGENTS.md, GEMINI.md, `.github/copilot-instructions.md`, `.cursor/rules/*.mdc`, or `.claude/rules/*.md` — then edits one copy. The copies silently disagree, and different assistants get different instructions.

The signal is precise:

- **Identical sections never fire** — exact copies are intentional sync.
- **Near-identical sections fire** — `similarity-threshold <= ratio < 1.0` (default 0.8) after normalization (lowercased word stream, code blocks stripped, heading text excluded so a retitled-but-identical section still reads as identical).

```
CLAUDE.md:14: Section 'Testing' is 87% similar to 'Testing' in
.github/copilot-instructions.md:13 — likely a drifted copy; reconcile
them or make them identical  [content-instruction-drift]
```

## Design

- Sections are sliced from the markdown AST (`markdown.headings()`), leaf-section granularity; sections under `min-section-words` (default 40) are skipped.
- One violation per drifted pair, deterministically anchored on the later section in `(path, line)` order, referencing the earlier copy with repo-relative path and file line.
- **Generated-file guard** (`ignore-generated`, default true): files carrying a "generated by … do not edit" marker are skipped — apm-compiled CLAUDE.md / copilot-instructions.md are near-copies *by design*, not drift. (skillsaw's own repo is the motivating case.)
- Perf: word lists and lengths are computed once per file; pair comparison is staged cheapest-first — difflib's exact `2*min/(min+max)` length bound (results-identical to `ratio()`, no false skips), then `quick_ratio()`, then `ratio()`; `SequenceMatcher` seq2 caching is exploited by keeping the anchor fixed per inner loop.
- `INFO` severity, `enabled: auto`, `since: 0.17.0` (version-gated so existing pinned configs are unaffected), no autofix — reconciling drift is a human judgment call.
- Neighbors, for the curious: `content-contradiction` (hardcoded contradictory phrase pairs), `content-inconsistent-terminology` (term variants), `content-redundant-with-tooling` (instructions duplicating tool config). None detect cross-file near-duplication.

## Tests

- 12 unit tests: detection with exact anchor/line/message assertions, identical-skip, below-threshold, min-words, single-file, code-blocks-don't-participate, generated-marker skip (and un-skip via config), config validation, custom threshold.
- Integration: realistic fixture (`tests/fixtures/content/instruction-drift/`) where CLAUDE.md's Testing section gained a sentence relative to the copilot-instructions copy; exact file/line assertions; silence check on a clean fixture; registered in the `BROKEN_FIXTURES` every-rule-fires-somewhere guard.

## Validation

- `make test`: 2482 passed; `make lint` clean; `make update` clean (docs regenerated in the second commit)
- Self-lint: 0 findings on skillsaw's own repo — verified 0 even with `ignore-generated: false` (compiled CLAUDE.md currently matches its `.apm` sources exactly)
- `openshift-eng/ai-helpers`: default run exit 0 (rule is since-gated by their pinned version); force-enabled across 40 plugins / 103 skills → **0 false positives**
- Perf on ai-helpers (5 runs): -1.0% delta with the rule enabled — noise

## Real-world sweep

Swept 169 public repos (curated list + GitHub code search for CLAUDE.md/AGENTS.md), fetching only their instruction files via the API and running this branch over each. 67 repos had ≥2 instruction files to compare; **5 had drift, 62 were clean, 0 errored**:

| Repo | Findings |
|---|---|
| **microsoft/TypeScript** | `Required behavior for coding agents` — AGENTS.md vs `.github/copilot-instructions.md`, 92% |
| **denoland/deno** | `High Level Overview` (92%) and `Common Patterns` (90%) — CLAUDE.md vs `.github/copilot-instructions.md` |
| 1amageek/SwiftAgent | `AgentTools` — CLAUDE.md vs AGENTS.md, 96% — **intentional per-assistant divergence**, see below |
| FreePeak/LeanKG | 4 findings across the CLAUDE.md / AGENTS.md / GEMINI.md triple |
| j5ik2o/oni-comb-rs | Japanese-language design notes drifted across CLAUDE/AGENTS/GEMINI, 92% (rule behaves sensibly on CJK text) |

Hand-verified the top two as genuine drift:

- **deno** is the textbook case: two hand-maintained copies of the same architecture overview where someone copy-edited only one — "user-visible" → "user visible", "including all of the" → "including all", an em-dash swapped for an en-dash, paragraphs re-wrapped. Any future change to the architecture description goes stale in one copy silently.
- **TypeScript** drifted by addition: AGENTS.md's copy of the required-behavior section gained a sentence pointing at copilot-instructions.md that the original lacks, while the shared bullet list underneath stays duplicated.

Hand-checked SwiftAgent too: its two files are 1005-line clones whose only differences are assistant-name substitutions (“Claude Code スタイル” vs “Codex スタイル”) — *intentional* per-assistant divergence, not drift. The identical-skip already absorbed the other ~30 cloned sections; for the one that differs on purpose, the standard inline suppression is the recipe (now documented in the rule doc, with tests locking in that the directive itself is invisible to the comparison — HTML comments and whitespace are blanked before sections are compared):

```markdown
<!-- skillsaw-disable content-instruction-drift -->
## Tool naming
...harness-specific content...
<!-- skillsaw-enable content-instruction-drift -->
```

The clean list is equally important for the noise profile: openai/codex, cline/cline, microsoft/vscode, oven-sh/bun, zed-industries/zed, google-gemini/gemini-cli, and 56 others produced zero findings. A ~7% hit rate with verified true positives supports shipping this auto-enabled at INFO.


stbenjam:
Implement the optional items from the review above

Issue #284, opened 2026-06-10
Migrate all markdown parsing to a Markdown AST (markdown-it-py); fixes splice at token spans


stbenjam:
## Why

Every piece of markdown structure in skillsaw — links, code fences, inline code spans, headings, HTML comments — is derived by per-line regexes scattered across ~17 content rules, suppression-directive parsing, and the autofix paths. Fixes re-locate their targets by string search (`line.find()` / `str.replace()`) instead of using the position the check matched.

A quality audit showed where that leads. The headline case, at SAFE confidence (plain `skillsaw fix`, no `--suggest`):

```bash
mkdir -p /tmp/a/docs && printf '# S\n' > /tmp/a/docs/setup.md
printf '# Agents\n\nBackup docs/setup.md.bak and docs/setup.md too.\n' > /tmp/a/AGENTS.md
skillsaw fix /tmp/a
# AGENTS.md is now: Backup [docs/setup.md](docs/setup.md).bak and docs/setup.md too.
#   -> wrapped 13 chars INSIDE the .bak token; left the real reference bare
skillsaw fix /tmp/a   # second run modifies the file again — not idempotent
```

#283 is the same bug class in `agentskill-rename-refs` (substring corruption, compounds per run). Other audit findings: cross-paragraph stray backticks hide broken links (a DOTALL regex blanks across paragraph boundaries, which CommonMark never does); indented code blocks are scanned as prose (the regexes don't know they exist); suppression directives inside fenced code blocks are honored; the broken-link fix drops anchors and silently skips titled links.

**Direction (agreed with maintainer): parse Markdown into a real token tree, exposed on the lint tree per the project rule "everything must be part of the parse tree". AST for reading; surgical span splices for writing. Never render the AST back to markdown** — round-trip rendering reformats whole files, violating the scope-the-fix autofix rule.

**Parser: `markdown-it-py>=3`** — pure Python, strict CommonMark, the engine under mdformat/MyST, no native wheels (safe for pipx and pre-commit distribution). New runtime dependency (current runtime deps are only PyYAML + ruamel.yaml). No mdit-py-plugins: frontmatter continues to be split by the existing `utils.parse_frontmatter` machinery.

## Approach: migrate first, then re-test

[...]

PR #428, opened 2026-07-16
Fix content-inconsistent-terminology: exclude headings, report line numbers

### LINKED ISSUE #427: content-inconsistent-terminology: word-boundary regex matches path segments and headings, not prose choices
## What happened

While cleaning up `content-inconsistent-terminology` violations in a real repo, I found the remaining hits (after fixing genuine prose inconsistencies) were not fixable without actively harming the docs -- the rule was matching directory names and markdown headings, not actual terminology choices.

## Root cause (traced in `rules/builtin/content/inconsistent_terminology.py`)

The rule runs plain `\bword\b` regexes (e.g. `\bcodebase\b`, `\bpull\s+requests?\b`) against the full raw body of each content file, then flags any file that matches a "minority" pattern -- even once, even alongside heavy use of the majority term, with no location/line info returned.

Two concrete false-positive sources from this:

1. **Path segments inside inline code spans read as standalone words.** A file containing a backtick-wrapped path like `` `.planning/codebase/CONVENTIONS.md` `` gets flagged as "using" the term `codebase` (per the `repo/repository/codebase` group), because `\bcodebase\b`'s word boundaries are satisfied by the surrounding `/` characters. The file may use `repo` dozens of times and never once use "codebase" as an actual noun -- it still gets flagged as inconsistent, purely because of an unrelated directory name.

2. **Headings and frontmatter descriptions are compared against body prose as if they were the same register.** A skill file titled `# Create Pull Request` (a completely normal, spelled-out heading) that consistently uses "PR" in its body prose gets flagged under the `PR/pull request/merge request` group, because the heading text matches the "pull request" pattern. Same issue for `function/method`: a rubric file using "method" correctly as a Go-language term (`Create()`/`Update()` methods) gets flagged as inconsistent with unrelated files that use "function" as a verb ("capabilities... function") elsewhere in the repo -- the regex can't distinguish the noun from the verb, or the technical term from the generic one.

## Suggested fixes

- Strip inline code spans (single backtick `` `...` ``) from the text before matching, the same way fenced code blocks are presumably already excluded -- path-like strings in backticks shouldn't count as prose terminology usage.
- Exclude heading lines and frontmatter fields from the scan, or at minimum weight/report them separately from body prose, since headings and descriptions are a different register than running text.
- Report the specific matching line(s), not just the file -- with no location info, tracking down *why* a file was flagged (vs. a real prose inconsistency) required reading the rule's source and manually grepping every pattern.

## Repro

Any repo with:
- A path reference like `` `.planning/codebase/foo.md` `` alongside heavy use of "repo" elsewhere -> flags the `repo/repository/codebase` group.
- A skill/doc titled with a spelled-out term (e.g. `# Create Pull Request`) that uses the abbreviation everywhere else in body prose -> flags the corresponding group.

Happy to share the specific repo/config if useful, or open a PR if this is a fix you'd rather receive as a patch.

tchughesiv:
## Summary

`content-inconsistent-terminology` was comparing headings against body prose as if they were the same register, and only reported the file, not the offending line.

- A skill titled `# Create Pull Request` that consistently says "PR" in its body was flagged as inconsistent — the heading text alone satisfied the `\bpull\s+requests?\b` pattern. Heading lines are now blanked before scanning so only running prose counts as a terminology choice.
- Violations now carry a `line` pointing at the actual matching occurrence (via `block=`/`line=` on `self.violation()`), instead of only the file path, matching the location contract every other content rule follows.

The other false-positive source described in the issue — a path segment inside a code span (e.g. `` `.planning/codebase/CONVENTIONS.md` ``) reading as a standalone word — turned out to already be handled by `read_body(strip_code_blocks=True)` (added by the markdown-it-py AST migration, #284); added a regression test to lock that in since it wasn't previously covered for this rule.

## Changes

- `src/skillsaw/rules/builtin/content/inconsistent_terminology.py`: blank heading line ranges (via `MarkdownDoc.headings()`) before matching, precomputed once per file; attach the first matching line to each violation.
- `src/skillsaw/rules/docs/content-inconsistent-terminology.md` (+ generated `docs/rules/...`): document that only prose counts, not headings or code spans.
- Unit tests in `tests/test_content_rules.py` for the heading false positive, the code-span false positive, and line-number reporting.
- Integration fixture + test (`tests/fixtures/content/inconsistent-terminology-registers`, `TestInconsistentTerminologyRegisters`) reproducing the exact scenario from the issue end-to-end.

Closes #427

## Test plan

- [x] `make test` — full suite passes (2615 passed)
- [x] `make lint` — black formatting clean
- [x] `make update` — regenerated docs/example config, `skillsaw lint .` self-lint stays green
- [x] Cloned `openshift-eng/ai-helpers`, ran `skillsaw`, exit 0
- [x] Manually reproduced the issue's exact repro (heading + code-span path) before/after the fix

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminology consistency checks to ignore headings and inline code, including paths.
  - Violations now point to the precise source line where inconsistent terminology appears.
  - Reduced false positives when terminology variations occur only in headings or code spans.

- **Documentation**
  - Clarified that only running prose is evaluated for terminology consistency.


tchughesiv:
Self-audit follow-up: the first commit's per-file scan-body cache was keyed by `cf.path`, but `CodeRabbitContentBlock` and `PromptfooPromptBlock` can emit multiple content blocks sharing one file path (one per instruction fragment / prompt entry). That collapsed every fragment in a `.coderabbit.yaml` or promptfoo config onto whichever fragment's body was computed last — silently dropping earlier fragments from scanning and misattributing matches (and their reported line) to the wrong file/block.

Fixed by keying the cache by list position instead of path. Added regression tests for both block types. Verified against `openshift-eng/ai-helpers`'s real `.coderabbit.yaml` (which has 5 `instructions:` fragments) — the reported line for its terminology violation moved from an unrelated line to the actual matching prose.

Also ran `benchmarks/bench.py --compare` against a `main` baseline per the repo's performance guidelines: no regressions, and the rule is now ~35% faster since bodies are computed once per file instead of once per term group.

stbenjam:
Thanks so much, the changes here make sense to me.

stbenjam:
I'm traveling and have a few more things I want to merge before I cut a new release, should be early next week.

PR #461, opened 2026-07-26
Make slop gate the verdict, not just carry a label

not-stbenjam:
Follow-up to #460. This commit was pushed about a minute after #460 was
squash-merged, so it missed the merge.

## The gap

#460 gave the Slopinator a `BLOCKING` severity, but blocking in name only.
Three things on `main` today let a slop finding ship anyway:

1. `SKILL.md` still says **"Keep the panel advisory. It does not gate merge."**
   with no exception for slop.
2. `APPROVE` is "Set when no unresolved BLOCKING findings remain" — nothing
   prevents the arbiter from resolving a slop finding by filtering it as a
   style nit in the noise-filter step.
3. Part A residue inherits the prose clustering rule ("prefer clustering to
   isolated hits"), so a single review-history comment can read as below the
   reporting threshold.

## Changes

- **Carve slop out of the advisory stance.** The panel stays advisory on
  everything else; review-history residue and generated prose must not merge.
  An unresolved Slopinator `BLOCKING` rules out `APPROVE` on its own and holds
  the verdict at `REQUEST_CHANGES`.
- **Add an arbiter bias against softening slop.** It is not filtered as a style
  nit in Step 3, and not downgraded because the rest of the change is good.
- **Block on every Part A instance.** Residue is binary — a comment either
  describes the shipped code or the conversation that produced it — so the
  clustering rule now explicitly governs Part B prose only. One review-history
  comment is enough.
- Split the reporting-volume guidance out of "What NOT to flag", which the
  added sentence had pushed past the 500-token `content-section-length` limit.

The precision guardrails are unchanged: "What NOT to flag" still protects long
"why" comments, house style, domain vocabulary, version-scoped documents, and
comments a reviewer explicitly asked for.

## Scope note

The panel posts its verdict with `gh pr comment`, so `REQUEST_CHANGES` is a
disposition in a comment rather than a GitHub review state — this makes slop
gate the *verdict*, not the merge button. A hard gate would mean posting a real
review via `gh pr review --request-changes` plus branch protection, which is a
larger change to Step 6 and is deliberately not in this PR.

## Verification

- `make test` — 2965 passed, unchanged from `main`
- `make lint` — clean
- `.venv/bin/skillsaw lint .` — 0 errors, 0 warnings, grade A+, 6 info-level
  violations (the `main` baseline; the regenerated report card shows no drift)
- All three skill copies byte-identical

### DIFF
--- a/.agents/skills/skillsaw-review-panel/SKILL.md
+++ b/.agents/skills/skillsaw-review-panel/SKILL.md
@@ -24,8 +24,13 @@ Run **7 specialist reviewers + 1 arbiter** to review a skillsaw PR.
   Cheaper — codebase context is derived once and shared — but later
   specialists see earlier ones' file reads, which can bias them.
 
-Keep the panel **advisory**. It does not gate merge. It surfaces findings;
+Keep the panel **advisory** on everything except slop. It surfaces findings;
 the maintainer and PR author review and decide ship.
+
+**Slop is the exception.** Review-history residue and generated prose must not
+merge. A `BLOCKING` finding from the Slopinator gates the verdict: the
+disposition is `REQUEST_CHANGES` until the text is fixed, and no amount of
+agreement from the other specialists converts it into an `APPROVE`.
 
 ## Arguments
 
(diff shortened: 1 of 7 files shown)


stbenjam:
Meh, the current version is fine.

PR #467, opened 2026-07-27
Replace inline provenance guards with declarative rule scoping

not-stbenjam:
## What

The maintainer review's central criticism of this branch: ecosystem ownership is decided once (`RepositoryContext.provenance()`), but the provenance *question* was copy-pasted as inline guards across the rule files — "the data generalizes, the dispatch doesn't." This PR makes the dispatch declarative. Behavior is byte-identical.

### Form 1 — whole-content skips (8 rules, guards deleted)

`Rule.provenance_scope` names the ecosystem whose format conventions a rule enforces (`"claude"` on `command-frontmatter`, `command-naming`, `command-sections`, `command-name-format`, `agent-frontmatter`, `plugin-json-required`, `plugin-naming`, `marketplace-registration`; `None` = ecosystem-neutral, the default). Rules iterate targets through `Rule.scoped_find(context, NodeType)`, and the filtering happens in exactly one place: `RepositoryContext.in_format_scope()`, reading each node's `provenance_dir()` against the cached `PluginProvenance` claim set.

- `PluginNode.provenance_dir()` is its own path; `CommandBlock`/`AgentBlock` return `path.parent.parent` — the exact directory the deleted guards probed, correct wherever the block hangs (PluginNode, CodexPluginNode, or the tree root for a root-level Codex plugin).
- Dual-manifest and unclaimed directories stay in scope; Codex-only directories are exempt — the same line the guards drew.

### Form 2 — conditional strictness (2 rules, one shared spelling)

`hooks-json-valid` and `mcp-valid-json` keep their semantic (ecosystem-tightened checks, never a skip) but now consult the one shared gate, `RepositoryContext.in_codex_only_plugin(path)`, replacing two local reimplementations of the owner-walk-plus-codex-only conjunction.

### Docs

The "Ecosystem provenance" section and the "Adding an ecosystem" recipe in `.apm/instructions/development.instructions.md` (mirrors regenerated via `make update`) no longer mention visiting rule files: a new evidence probe in `provenance()` is honored by every scope-declaring rule the moment it lands.

## Verification

- `grep -rn "is_codex_only_plugin" src/skillsaw/rules/` — zero hits. The only remaining provenance call in rules is the symlink-containment walk in `agentskills/_helpers.py` (`codex_plugin_owning`), which asks a different question.
- `make test`: 3480 passed. `make lint` and `make update` clean; `src/skillsaw/rules/` is net −44 lines.
- Real-corpus canary (`skillsaw lint ~/git/plugins --format json`): 973 violations before and after, identical per-rule counts. Fresh shallow clone of `openshift-eng/ai-helpers`: output identical modulo the wall-clock line (its one pre-existing error — their `docs/` drift — appears identically in both runs).
- Mutation sanity: inverting the dual-manifest handling in `in_codex_only_plugin` fails 3 `TestDualManifestBackwardCompat` tests; inverting `in_format_scope`'s dual handling fails the new `TestProvenanceScopeMechanism` pins plus an existing provenance test. Both mutations reverted.
- New tests in `tests/codex/test_provenance.py` (`TestProvenanceScopeMechanism`) pin which rules declare the scope, the block- and PluginNode-level filtering, and the neutral default.


not-stbenjam:
@codex review

not-stbenjam:
Re Qodo's rule-violation finding: `AgentFrontmatterRule` declares no `repo_types` on main either — its activation is deliberately block-discovery-driven (no agent blocks in non-Claude repos means nothing to check), and this PR only adds a `provenance_scope` line without touching activation semantics. Pre-existing design, out of scope here.

PR #489, opened 2026-08-07
fix: escape JavaScript line terminators in generated docs

### LINKED ISSUE #488: docs: escape line terminators in generated escJsAttr
## Kind of issue

Other

## skillsaw version

0.18.0

## Reproducer

Run the docs generator:

```sh
skillsaw docs -o docs/ --theme crimson-red
```

The generated `docs/index.html` contains:

```js
function escAttr(str) {
  if (!str) return '';
  return String(str)
    .replace(/&/g, '&amp;')
    .replace(/"/g, '&quot;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;');
}

function escJsAttr(str) {
  return escAttr(String(str).replace(/\\/g, '\\\\').replace(/'/g, "\\'"));
}

var value = 'first\nsecond';
var html = '<button onclick="navigateTo(\'' + escJsAttr(value) + '\')">Open</button>';
```

## Expected behavior

`escJsAttr` should escape JavaScript line terminators before the value is placed in an inline event-handler attribute. The generated handler should remain valid JavaScript and preserve the value. This should include CR, LF, and ideally U+2028/U+2029.

## Actual behavior

Backslashes and single quotes are escaped, but CR/LF are returned unchanged. A value containing a newline can produce an inline handler containing a raw line terminator inside a single-quoted JavaScript string, causing the handler to fail to compile when the generated HTML is parsed.

## Configuration

Default settings; no special configuration is required.

## Additional context

This was observed after regenerating the `docs/index.html` in [openshift-eng/ai-helpers#685](https://github.com/openshift-eng/ai-helpers/pull/685). `escAttr` handles the HTML attribute context, but `escJsAttr` also needs to handle the nested JavaScript-string context. A likely fix is to escape line terminators in `escJsAttr` before calling `escAttr`.

stbenjam:
## Summary

Generated documentation builds inline event handlers with JavaScript string values nested inside HTML attributes. The old helper manually escaped a single-quoted JavaScript literal and omitted ECMAScript line terminators, which could leave invalid handlers in generated pages.

Use `JSON.stringify` as the JavaScript string-literal serializer, escape U+2028 and U+2029 for compatibility, then HTML-escape the resulting double-quoted literal. Update all inline-handler call sites to use the complete serialized literal.

Fixes https://github.com/stbenjam/skillsaw/issues/488

## Validation

- `.venv/bin/pytest tests/codex/test_docs_output_safety.py -q` — 65 passed
- `make test` — 3903 passed, 22 warnings
- `make lint` — passed
- `make update` — passed with no generated changes
- `git diff --check` — passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safety and reliability for generated navigation and search links containing special characters.
  * Correctly handles quotes, backslashes, line breaks, Unicode separators, and HTML-sensitive characters.

* **Tests**
  * Expanded coverage to verify safe JavaScript serialization across supported special characters.


src/skillsaw/docs/html_renderer.py
```diff
@@ -1201,14 +1202,23 @@ def _get_js() -> str:
...
-    // handler compiles, so the JS escapes must survive that decode: \' stays
-    // \', while " arrives as a plain quote which cannot close a
-    // single-quoted JS string.
-    return escAttr(String(str).replace(/\\\\/g, '\\\\\\\\').replace(/'/g, "\\\\'"));
+    // escJs returns a complete double-quoted JS string literal. HTML-escape
+    // that literal so it can be placed inside a double-quoted attribute;
+    // innerHTML decodes the entities before the inline handler compiles.
+    return escAttr(escJs(str));
   }
+  // END_ESCAPERS
```
stbenjam: Why this extra layer of indirection lol

PR #521, opened 2026-08-28
Add content-broken-external-reference: an opt-in dead-link check for external URLs

stbenjam:
`content-broken-external-reference` is the outward-facing sibling of `content-broken-internal-reference`. It pulls every `http(s)` link out of the markdown AST, asks each server whether the page is still there, and reports the ones that are definitively gone.

It is also the first rule in skillsaw that opens a network connection, which is what the rest of this PR is about.

## The conservative design

Two constraints shaped everything.

**1. A lint run stays hermetic unless you say otherwise.**

`default_enabled = False` — opt-in, never `auto`. No repository type, no detected format, and no `enabled: auto` cascade can start making requests on a user's behalf. Turning it on takes an explicit `enabled: true` in `.skillsaw.yaml`, any other setting for the rule, or `--rule content-broken-external-reference` on the command line.

The rule docs push users toward `--rule` in a scheduled workflow rather than a per-PR gate, and `docs/ci.md` ships the recipe.

**2. Only definitive evidence is a violation.**

| Response | Verdict |
|---|---|
| `404`, `410` | **Violation** |
| `401`, `403`, `429` | Not a violation — bot walls and rate limits describe the runner, not the link |
| `5xx` | Not a violation — the origin is having a bad day |
| Timeout, DNS failure, refused connection, TLS error | Not a violation — the network between runner and host |
| Redirect past the hop cap, or leaving `http(s)` | Not a violation |

A linter that fails CI because a documentation site rate-limited the runner is worse than no linter at all — it teaches people to stop reading the output. Nothing in the inconclusive column is even reported at info level; it is silent.

**Mechanics.** `HEAD` first; one `GET` retry on `405`/`501`, reading status and headers and never the body. Redirects capped at five hops, and a redirect out of `http(s)` ends the chain rather than handing a repository-authored string to a non-HTTP client. URLs carrying credentials (`https://user:token@host/…`) are skipped entirely rather than put on the wire. URLs are de-duplicated repo-wide before any I/O — forty skills citing one URL cost one request — and every occurrence still gets its own file and line. A per-request `timeout` (5s) and a run-wide `total-budget` (30s) bound the run; when the budget runs out, the URLs still queued are never requested and the run emits exactly **one** info-level notice saying how many, never one per URL. Requests run on a small `ThreadPoolExecutor` over stdlib `urllib` — **no new dependency** — identifying as `skillsaw/<version>`, with a per-thread opener so nothing is shared across workers.

`supports_autofix` is false. There is no mechanical fix for a dead URL.

Options: `timeout`, `total-budget`, `concurrency`, `ignore` (glob when it contains `*`/`?`/`[`, literal prefix otherwise). Methods are not configurable.

## Test plan — nothing touches the real internet

`tests/test_external_links.py` (38 tests) runs everything against an `http.server` bound to `127.0.0.1` on an ephemeral port, scripting `200`/`403`/`404`/`410`/`429`/`500`, a redirect to a 404, a redirect loop, a redirect into `ftp://`, a HEAD-rejecting route, and two routes that never answer in time. The server records every hit, so the tests assert on what was requested *and* on what was not — that `/missing` is requested exactly once for three occurrences, that a URL inside a code fence is never requested, that `/ok` sees `HEAD` and only the 405 route earns a `GET`, that ignored URLs generate no traffic at all.

The no-real-network guarantee has three legs:

1. **Nothing outbound exists to hit.** Every URL in the suite is `127.0.0.1:<ephemeral>`. The offline test points the fixture at a *closed* port and asserts zero violations; the DNS test uses `host.invalid` (RFC 2606, never resolves).
2. **A default run makes no requests.** With the fixture's opt-in config removed, a full CLI lint leaves `server.hits == []`.
3. **A default run cannot make one.** The stronger test patches `socket.socket.connect`/`connect_ex` to raise and runs the entire default rule set in-process. Any rule reaching for the network fails loudly instead of quietly succeeding against a reachable host.

An autouse fixture also clears `http_proxy`/`https_proxy`/`ALL_PROXY` and sets `no_proxy=*`, so a developer machine or runner with an ambient proxy cannot push these loopback requests off-box.

Markdown content lives in `tests/fixtures/content/external-links{,-slow}/` per the repo's fixture-first rules; only the port — unknowable until the server binds — is substituted into the copy.

`TestRuleCoverage` gains a documented `NETWORK_RULES` exemption (a rule whose verdict depends on a live response cannot fire from a static fixture) plus a companion test that fails if an exempted rule has no local-server coverage, so the exemption cannot rot into a hole.

## Validation

- `make test` — **4706 passed**, 0 failed. New rule at 92% line coverage.
- `make lint` — clean. `make format` applied.
- `make update` — idempotent; regenerated `.skillsaw.yaml.example`, `docs/rules/content-broken-external-reference.md`, the content-intelligence index, and `docs/research.md`. `make verify-apm` OK. Self-lint: 0 errors, 0 warnings, badge/card unchanged (the rule does not fire on skillsaw's own content).
- `make benchmark-save` on `origin/main`, `make benchmark-compare` on this branch — **no regressions detected**. Expected: the rule is disabled by default, so it never runs in the benchmark.
- `openshift-eng/ai-helpers` — `skillsaw lint` exits **0**. Re-run in-process with `socket.connect` patched to raise: `exit=0 connect_attempts=0`.

## Deviations from the spec

- **`403`/`429` are fully silent**, not "at most informational". The spec allowed an info-level report if an existing pattern fit; none did, and an unactionable info line on every bot-walled link would be noise the user cannot fix.
- **`total-budget: 0` disables the cap** rather than meaning "no time at all". A scheduled job with thousands of links needs an unlimited setting more than it needs a way to spell "check nothing".
- **Files under `template/`, `templates/`, `_template/` are skipped**, matching `content-broken-internal-reference`. Placeholder targets there are intentional. Not in the spec; documented on the rule page.
- **URLs with userinfo are skipped**, not just redacted in messages. A linter should not put an author's pasted credentials on the wire.
- **Bare URLs in prose are out of scope**, as the spec permitted. `text_segments()` would have needed its own URL grammar, and a bare URL in prose is text, not a reference.
- `since = "0.20.0"`, which matches the unreleased in-tree version — no version bump needed.

Assisted-by: Claude:claude-opus-5

https://claude.ai/code/session_01J1Qt3WNQCWATbFrjxLQZPL

## Summary by CodeRabbit

* **New Features**
  * Added opt-in detection for broken external HTTP(S) links returning 404 or 410.
  * Added network controls for linting, fixing, baselining, and badge generation, including offline mode and private-host access.
  * GitHub Actions now disable network access by default, with a configurable override.

* **Documentation**
  * Documented external-link checking, network safety controls, configuration, and scheduled CI usage.
  * Updated rule indexes and security guidance.

* **Bug Fixes**
  * Prevented network-dependent rules from being automatically selected in changed-rule workflows.


stbenjam:
## Vetting the real-repo findings

I verified every finding the impact report attributes to `content-broken-external-reference` on **prodsec-skills**, plus the ten listed lines on **awesome-copilot** and both on **ai-helpers**, by cloning each repo at the reported ref, extracting the exact URL from the reported line, and probing it with `curl` — HEAD and GET, with the rule's own User-Agent and with a browser one, to separate genuine 404s from bot walls.

**One systematic false-positive class, and it is real.**

### prodsec-skills @ `b55c6b9`

| # | Line | URL | HEAD | GET | Verdict |
|---|---|---|---|---|---|
| 1 | `ge-core/ai-systems-security/SKILL.md:259` | `github.com/…/blob/main/module/skills/container-hardening/SKILL.md` | 404 | 404 | ✅ **True positive** — the skill lives at `ge-core/container-hardening/`, not `module/skills/`. Repo root is 200, so not a bot wall. |
| 2 | `apache-camel-security/SKILL.md:45` | `camel.apache.org/components/4.0.x/others/shiro.html` | 404 | 404 | ✅ **True positive** — the whole `4.0.x` doc tree is retired (`/components/next/others/shiro.html` is 200). |
| 3 | `apache-camel-security/SKILL.md:46` | `camel.apache.org/components/4.0.x/others/spring-security.html` | 404 | 404 | ✅ **True positive** — same retired tree. |
| 4 | `defense-in-depth/SKILL.md:73` | `nvlpubs.nist.gov/…/NIST.SP.800-53r5.pdf` | **404** | **200** | ❌ **FALSE POSITIVE** |
| 5 | `graphql-security/SKILL.md:40` | `github.com/stems/graphql-depth-limit` | 404 | 404 | ✅ **True positive** — repo gone; 404 for a browser UA too. |
| 6–10 | `semgrep-rule-creator/SKILL.md:163,165,166,167,168` | 5× `raw.githubusercontent.com/semgrep/semgrep-docs/…/*.md` | 404 | 404 | ✅ **True positive** ×5 — semgrep-docs renamed these to `.mdx`; the same paths with `.mdx` return 200, and line 164 (already `.mdx`) was correctly *not* flagged. |

**9 true positives, 1 false positive.** The rule is finding real link rot here — including a self-link the repo has an open `fix-links-*` branch for.

### awesome-copilot @ `f11a4e4` (10 listed lines sampled)

| URL | HEAD | GET | Verdict |
|---|---|---|---|
| `azure.microsoft.com/…/content-moderator/` | **404** | **200** | ❌ **FALSE POSITIVE** (same shape as #4) |
| `ai.google/responsibility/`, `ai.google/research/responsible-ai/` | 404 | 404 | ✅ True positive ×2 (404 for browser UA too) |
| 4× `docs.microsoft.com/en-us/…` | 404 | 404 | ✅ True positive ×4 — pages retired in the move to learn.microsoft.com. Intermediate redirect behavior is unstable client-to-client, so a couple are also **environment-sensitive**. |
| `docs.wandb.ai/guides/prompts` | 404 | 404 | ✅ True positive |
| `azure.github.io/Azure-Verified-Modules/contributing/terraform/testing/` | 404 | 404 | ✅ True positive |

### ai-helpers @ `18631f3` (both findings)

| URL | HEAD | GET | browser GET | Verdict |
|---|---|---|---|---|
| `github.com/openshift/chai-bot` | 404 | 404 | 404 | ✅ True positive |
| `olm.operatorframework.io/docs/tasks/creating-catalog-from-index/` | 404 | 404 | 404 | ✅ True positive |

*(No finding in any sample turned out to be a bot wall returning 404 to the runner — every 404 I could not confirm as genuine was the HEAD/GET class below.)*

---

## Root cause

**HEAD/GET divergence.** RFC 9110 says a `HEAD` response must be what `GET` would return minus the body. A meaningful minority of servers — CDNs and WAFs especially — do not comply and answer `404` to `HEAD` while serving the resource on `GET`. The rule probed with `HEAD` and retried with `GET` only on `405`/`501`, so a mis-implemented `HEAD` became a reported 404.

That is precisely the ambiguous case this rule's design says it must stay silent about, so it is a bug in the rule, not in the repos.

## Fix — `64b72c6`

A status that would become a violation is now **re-asked with `GET`, and `GET`'s answer is the one that counts**. If `GET` says anything other than `404`/`410` — including nothing at all, because the budget expired mid-confirmation — the link is not reported.

- Cost: one extra request **per candidate violation only**. Healthy links still cost a single `HEAD`; there is a test pinning that (`test_healthy_links_cost_a_single_head`).
- All 9 prodsec-skills true positives survive the confirming `GET`, so nothing real is silenced.

Regression coverage against the local-server harness reproduces the exact failure shape — `/head-404-get-200` and `/head-410-get-200` routes that answer 404/410 to `HEAD` and 200 to `GET` — plus a test that an unconfirmable 404 stays silent. I verified the new tests fail without the fix (5 failures) and pass with it.

`make test` 4753 passed · `make lint` clean · `make update` clean.

Assisted-by: Claude:claude-opus-5

https://claude.ai/code/session_01J1Qt3WNQCWATbFrjxLQZPL

stbenjam:
@coderabbitai review

stbenjam:
## Response to the review panel (round 1)

All four **Required Actions** are done, plus most of the optional follow-ups. Nothing was declined outright; two items were resolved differently than proposed and are called out below.

### Required actions

**1. Operator-side network gate** — `7337083`

`--no-network` / `SKILLSAW_NO_NETWORK` on `lint`, `fix`, `baseline`, and `badge`, mirroring `--no-custom-rules`. It is **declarative**: `Rule.requires_network` is a base-class attribute, the gate filters on it after *every* loader (builtin, plugin, custom), and it runs independently of `--rule` — selecting which rules run is not the same as granting network access. `action.yml` gets `no-network` defaulting to `'true'`. A `NetworkAccessWarning` renders on stderr through the same path as `CustomRuleWarning`:

```
⚠ Network access enabled for: content-broken-external-reference (use --no-network to skip)
```

`_port.py` hardcodes `no_network=True` alongside its existing `no_custom_rules=True`. README and `docs/ci.md` now state the guarantee in terms of the gate rather than the class default.

I also took the Security Reviewer's host-restriction item, which the synthesis left out of the required list: destinations are confined to public hosts (loopback / private / link-local / reserved / multicast / IPv4-mapped / `localhost` names refused unless `allow-private-hosts: true`), which closes the `169.254.169.254` oracle concretely. **And the same admission gate — including `ignore` — now re-runs on every redirect hop**, so `ignore` is no longer bypassable by a 302 and userinfo cannot be re-sent (Security 4). Honest residual, documented: an internal *hostname* that resolves privately is not caught without a resolver lookup, and DNS rebinding is out of scope — `--no-network` is the control that does not depend on parsing.

**2. Clamp and sanitize every option** — `7337083`, `ebffa6e4`

`_MAX_TIMEOUT = 30`, `_MAX_TOTAL_BUDGET = 600`, `_MAX_CONCURRENCY = 32`, named in the `_MAX_REGEX_TIMEOUT` style and interpolated into the schema descriptions so the docs cannot drift. Negative and non-finite values fall back to the **default**, not to the clamp ceiling and not to "no cap" — `timeout: .inf` + `total-budget: 0` no longer yields an unbounded socket timeout. `OverflowError` is handled, `bool` is rejected for numeric options, and `ignore` gets an `_as_list` guard so `ignore: "https://…"` can no longer iterate into `startswith("h")` and silence every URL. `TestUntrustedConfig` parametrizes each hostile value.

On the soft-budget finding (Python 1 / Security 3): `_probe_all` now collects with `concurrent.futures.wait` against a deadline and exits through `pool.shutdown(wait=False, cancel_futures=True)`, so a stuck worker cannot hold the run; anything unresolved is counted as unchecked. With `timeout` clamped and redirects capped at 5, a worker's worst case is bounded rather than unbounded.

**3. `THREAT_MODEL.md` and `REVIEW.md`** — `7337083`

§1 scoped, `external_link_targets` entry point added, **T18** added covering SSRF / internal-network probing / beaconing with the implemented controls and an explicit residual-risk note, §7 `updated:` line recorded. `REVIEW.md:71` no longer tells reviewers skillsaw makes no network calls — it names the one exception and requires justification for widening it.

**4. The recipe can now fail** — `7337083`

Both copies get `--strict` and `-v`, with prose saying why (`warning` severity vs `fail-on: error`, and that `-v` is what surfaces the budget notice). Also brought in line with the file's own conventions: `checkout@v5`, `permissions: contents: read`, `persist-credentials: false`, consistent pinning, and a `no-network` row in the inputs table.

### Optional follow-ups taken

- **The tripwire test now tests something.** `test_default_run_cannot_open_a_socket` asserts on a recorded `attempts` ledger *and* on the absence of `rule-execution-error` in the returned violations, and also patches `getaddrinfo`. You were right that it could not fail as written.
- **`NETWORK_RULES` guard** replaced with an assertion on observed firing against the local server — no more substring match.
- **`except Exception` narrowed** to the network families (and further in `ebffa6e4`: bare `ValueError` removed, since `http.client.InvalidURL` is an `HTTPException` anyway).
- **HEAD-only 404 confirmed with GET** — this one shipped before the panel, in `64b72c6`; it was found independently by vetting the rule-impact report, where 2 of the findings I could verify were live pages (`nvlpubs.nist.gov`, `azure.microsoft.com`) answering 404 to HEAD.
- **Budget notice** made baselinable: the varying count is out of the message (which is what `fingerprint_violation` hashes for a file-less violation) and goes to the log instead; carries a `fingerprint_discriminator`.
- **`changed-rules.py`** filters `requires_network` rules, so `rule-impact.yml` stops force-running this against seven third-party repos on every `content/` PR.
- **Tests**: hop-cap test with a chain of *distinct* hops (`/redirect-loop` only tripped urllib's untouched `max_repeats`, as you noted); `501` parametrized alongside `405`; template-directory skip; image/badge destinations; `_request_url` `ValueError` branch; the two enablement behaviours (a bare setting enables; `enabled: false` keeps it off) and `--rule` force-enabling; liveness assertions on the three vacuous tests; `_closed_port` converted to a fixture that holds the socket bound (TOCTOU under xdist); the `host.invalid` DNS test now patches `getaddrinfo` and stays on loopback — it was the one test leaving the machine.
- **Docs**: a "What is sent, and to whom" section (IP, **full path and query string** — the userinfo skip does not cover `?access_token=…` — UA string, and that `HTTP_PROXY`/`NO_PROXY` are honored); that inconclusive answers are silent so a clean run is not proof; that the budget notice needs `-v`; that any non-`enabled` key activates the rule; that config-enabling makes the grade and badge network-dependent.
- **Prose**: both "worse than no rule at all" closers deleted, the six polish items, and the two fixture spots — the bare autolink now sits in prose where a naked URL is natural, and `/definitely-not-requested` is renamed `/v1-chargebacks`.
- **`.apm/instructions`**: a "Network-touching rules" entry. Note it is deliberately *short* — the full invariant list lives in the rule doc, because inlining it pushed `development.instructions.md` over skillsaw's own 4,000-token instruction budget and tripped `content-progressive-disclosure`. Applying the rule's own advice seemed better than suppressing it.

### Resolved differently

- **`action.yml` `rule` input** (Ecosystem 2): not added. Adding a general argument-passthrough input to the Action is a larger interface decision than this PR should make unilaterally, and the recipe works today via the CLI. Worth a follow-up issue rather than a rushed input here.
- **Scope-creep guardrail** (Ecosystem 4): covered implicitly by the new "Invariants for a future network rule" section on the rule page rather than as separate prose.

### Validation

`make test` **4817 passed** (106 in `test_external_links.py`, up from 38) · `make lint` clean · `make update` clean tree · `make benchmark-compare` no regressions · `openshift-eng/ai-helpers` exits 0 with `connect_attempts=0` under a socket-bricked run.

Assisted-by: Claude:claude-opus-5

https://claude.ai/code/session_01J1Qt3WNQCWATbFrjxLQZPL

stbenjam:
I will note I am skeptical of this rule.  As it is opt-in, its not very discoverable.  Perhaps the on-board skill should ask and offer to setup a weekly github action or something.

Really, there's a lot of tools for markdown link checking.   So I'm torn, it legitimately produces bad skills to have broken links.

stbenjam:
## Round 2 response

All seven required actions are addressed. Detail below, then what else changed.

### 1. `allow-private-hosts` is no longer a repo setting

The key is gone from `config_schema` entirely. Reaching non-public hosts is now
`--allow-private-hosts` / `SKILLSAW_ALLOW_PRIVATE_HOSTS`, threaded through
`Linter` the way `no_network` is and pushed onto the rule in
`_apply_network_gate`. `.skillsaw.yaml` has no say: a repo writing
`allow-private-hosts: true` lands an unrecognized option that changes nothing,
which `test_a_repo_config_key_cannot_grant_private_access` pins along with an
assertion that the key is absent from the schema.

The two environment variables deliberately resolve a typo in opposite
directions — `_env_restriction` (takes capability away, so `=disabled` gates)
and `_env_permission` (hands capability out, so anything but an explicit yes
withholds it). Both spellings are parametrized.

The loopback test server is now reachable only because the *operator* opted in:
`_run_rule` takes a constructor argument, and the CLI tests pass the flag. The
fixture `.skillsaw.yaml` files no longer carry the key.

### 2. Alternate IPv4 literals

`_as_ip_literal` asks `ipaddress` first and falls back to
`ipaddress.ip_address(socket.inet_ntoa(socket.inet_aton(host)))` under
`OSError`/`ValueError` — the liberal parser gets the final say on what counts as
a literal, and it still resolves nothing. `2852039166`, `2130706433`,
`0x7f000001`, `0177.0.0.1`, `127.1`, `0x7f.0.0.1` and `0` are all refused, on
the authored URL and on redirect hops.

`TestPrivateHostConfinement` parametrizes each spelling three ways: refused by
`_admit`, refused by `_accepts_hop`, and *admitted* under the operator's opt-in
— that last one is what stops a normalization that simply rejects anything
numeric from passing. Each case also asserts its own premise (`ipaddress` really
does raise, `inet_aton` really does yield the canonical address), so the
parametrize cannot rot into testing nothing. `[::]`, `[fd00::1]`, `127.0.0.1.`
and `LOCALHOST` are covered too.

### 3. The run is bounded

`total-budget: 0` no longer means unlimited — `0` and negatives both fall back
to the default, there is no value that disables the cap, and the
self-contradictory schema string is rewritten.

`ThreadPoolExecutor` is gone. Probes run on `daemon=True` threads over a
deque, joined against `deadline + timeout`; anything still running is abandoned,
counted as unchecked, and cannot hold the interpreter at exit.
`test_workers_that_never_return_are_abandoned` blocks `_probe` outright and
asserts the elapsed bound, `test_probe_threads_are_daemons` asserts the daemon
flag, and a worker exception is re-raised from `_probe_all` so a rule bug still
reaches `linter.py:1086` rather than reading as "nothing to report".

While there: each redirect hop gets the remainder of the caller's window instead
of a fresh full `timeout`, and the intermediate body urllib reads before opening
the next hop is suppressed (`_EmptyBody`), so the uncapped `fp.read()` on the
GET-confirmation path is closed. `TestNoBodyIsEverRead` covers all five redirect
codes — the base class aliases them, so overriding 302 alone was not enough —
and I verified those tests fail without the fix.

### 4. T18

Blank line deleted. Rendered through the project's own markdown-it: T18 now
appears in the row list alongside T1–T17 rather than as a paragraph. The
controls cell describes what the code does after this round, and §7's entry is
back in chronological order.

### 5. The CI recipe

There is now one copy, in `docs/ci.md`. The rule doc links to it and says why
there is only one. `skillsaw explain` no longer teaches a permanently green job.

### 6. Review-history residue

`_as_list`'s "used to iterate" is a present conditional. The "first real-repo
run" sentence in `_CONFIRM_WITH_GET` is deleted. `TestHeadIsNeverEnoughToConvict`
states RFC 9110 and the two hosts as standing facts, and `:867` reads "A link
that really is gone is still reported after confirmation." I also caught one
more of the same shape that landed in the fix itself — a comment saying `0`
"used to mean unlimited"; the rule is new, so it never did.

The `_parser.py` docstring is reduced, the banner/class ordering is fixed, the
fixture paths name resources rather than assertions, and `action.yml`'s
`no-network` description now matches its `no-custom-rules` sibling's register.

### 7. Behavioural gate coverage

`test_flag_gates_every_rule_executing_subcommand` parametrizes lint/fix/baseline/badge
and runs each twice: once reaching the server (liveness), once with the flag and
asserting zero hits. `test_env_var_gates_every_rule_executing_subcommand` does the
same for `SKILLSAW_NO_NETWORK` on all four rather than just `fix`.

I also took Architecture 2, which makes the guarantee structural rather than
test-enforced: `Linter.__init__` reads `SKILLSAW_NO_NETWORK` itself, so an
embedder or a future subcommand cannot bypass the operator's export by forgetting
to call the CLI helper. `test_the_env_var_reaches_a_directly_constructed_linter`
pins it. The custom-rule gate test now drives the real loader with a sentinel
file instead of calling `_apply_network_gate()` by hand.

Architecture 3 is in too: `--rule <gated rule> --no-network` raises instead of
exiting 0 over an empty rule set.

---

### Also, from the optional list

- **QA 2** — `_no_proxy` moved into `conftest.py` and grew teeth: a session
  autouse fixture refuses any `connect`/`connect_ex` to a non-loopback address,
  suite-wide. That is the guard the module docstring was promising. I verified it
  fires (off-machine connect raises, loopback still works).
- **Architecture 6 / Security 4** — `test.yml` and `rule-impact.yml` both export
  `SKILLSAW_NO_NETWORK` at job level, so a clone step added later inherits it.
  Environment rather than flag in rule-impact because its merge-base venv is an
  older skillsaw that would reject the argument.
- **Writer 4 / 5** — `docs/supply-chain-protection.md` gained a "Network access"
  subsection under "Skillsaw as a vector" and `--no-network` in its top warning;
  `docs/configuration.md`'s trusted-parent admonition now reads
  `--no-custom-rules --no-plugins --no-network`.
- **Writer 6 / Architecture 5 / Ecosystem 4** — the `../../` links in the APM
  source are inline code, so the six generated copies no longer point above the
  repository.
- **Ecosystem 3 / Writer 3 / 7 / 8 / 10 / 11** — "Deliberately not a general link
  checker", the notice-count wording, the Action's missing per-rule selector, the
  UA short form, the lower clamps, and the fragment-stripping `ignore` gap are all
  in the rule doc now.
- **Python 5** — `_clamp` is generic over a `TypeVar`, so the `concurrency` call
  site types as `int`.
- **QA 5 / 6 / 7 / 14** — the slow fixture's config is now read by an end-to-end
  CLI test of the budget notice (path-less INFO violation, exit 0, count on the
  `-v` log); the abandon-pending branch has its elapsed-time assertion; a rule bug
  propagating rather than being swallowed is pinned at both levels; and
  `_QuietServer.handle_error` only swallows the three disconnect errors.

### One thing this round found on its own

Merging `origin/main` brought in `skillsaw feedback`, which shells out to
`skillsaw lint` — a rule-executing path that arrived after the gate was built.
The child reads the linted repository's `.skillsaw.yaml`, so building a
diagnostic bundle would have started requesting every external URL the repo
links, inside a 120s budget. It now passes `--no-network` unconditionally;
unlike custom rules there is no `--with-extensions` escape hatch, since a
snapshot has no reason to go out on the network.

### Validation

`make test` 4923 passed · `make lint` clean · `make update` clean ·
`skillsaw` on `openshift-eng/ai-helpers` exits 0. `origin/main` merged before
push. I re-checked the hermeticity guarantee is not vacuous by running the same
socket-ledger technique with the rule enabled: 14 attempts recorded, versus 0 on
a default run.

Assisted-by: Claude:claude-opus-5

https://claude.ai/code/session_01J1Qt3WNQCWATbFrjxLQZPL

stbenjam:
## Round-3 response

All five required actions are landed in `cd7b876` and `2ce066d`. Optional follow-ups are deliberately untouched this round. `make test` (4947 passed), `make lint`, `make update` all clean; `openshift-eng/ai-helpers` exits 0.

### 1. Normalize the host the way the transport does, before classifying it

`_canonical_host()` applies `unquote()` then `.encode("idna").decode("ascii")` and returns `None` on `UnicodeError`/`ValueError`. `_is_public_host` classifies that string instead of the authored one, and `_request_url` rebuilds the netloc from it through `_canonical_netloc()` (port and IPv6 brackets carried through untouched), so what is requested, what `ignore` matches, and what gets reported are all the string the confinement classified.

Verified the bypass is real and closed, by neutralizing `_canonical_host` to the identity and re-running `_admit`:

```
without fix: ['http://169%2E254%2E169%2E254/latest/meta-data/', 'http://%31%32%37.0.0.1/x',
              'http://loc%61lhost/x', 'http://169。254。169。254/x', 'http://１２７.0.0.1/x',
              'http://①②⑦.0.0.1/x']
with fix:    [None, None, None, None, None, None]
```

Coverage: the six spellings from the verdict plus `０x７f.0.0.1` are in `_ALTERNATE_SPELLINGS`, so `test_non_canonical_ipv4_literals_are_refused`, `test_non_canonical_ipv4_redirect_hops_are_refused` and `test_the_operator_can_still_reach_them` all run over them. The liveness half of the first now computes the wire host from the stdlib — `unquote(host).encode("idna").decode("ascii")` — and asserts `inet_aton` on *that* yields the claimed address, so the parametrize still proves its own premise for both classes. `test_the_operator_can_still_reach_them` asserts `_admit(url) == url` with the host replaced by its wire spelling, which pins the netloc rebuild. `http://127%2E0%2E0%2E1/x` and `http://loc%61lhost/x` were added to `test_non_public_hosts_are_refused_by_default`, and `test_ignore_sees_the_url_the_transport_will_request` pins the `ignore` half.

Fail-closed on an unencodable host is now the documented behavior; T18's controls column and the rule doc's spellings section were rewritten to describe both classes rather than only the `inet_aton` one.

### 2. Stop `skillsaw fix` from sweeping the network once per autofix pass

Took the preferred option: `_run_fix` passes `no_network=True` unconditionally at both `Linter(...)` sites, with the reasoning in a comment at the first one. `fix` therefore makes zero requests on any pass, on either `Linter`, whatever the flags or `.skillsaw.yaml` say — so the `max_passes` multiplier is moot and `total-budget` stays the bound that matters for `lint`. The `--no-network`/`--allow-private-hosts` flags stay on the `fix` subparser so uniform CI argv keeps working.

The `fix` legs of `test_flag_gates_every_rule_executing_subcommand` and `test_env_var_gates_every_rule_executing_subcommand` pinned the old behavior, so `fix` is out of both parametrizes and `test_fix_never_probes_whatever_the_flags_say` replaces them — a stronger form of QA's suggested test: it runs `lint` first on the same repo, config and opt-in as a liveness check, then asserts `fix` leaves the ledger completely empty rather than at one hit per URL.

### 3. The last two pieces of review-history residue

- `:896` → `(0, 30.0),  # 0 is not "no cap" — it means the default`
- `:938` → "Iterating the string **would yield** single characters, and ``url.startswith("h")`` **would then match** every http(s) URL — the rule would report nothing while appearing to run."

Both reread before committing: neither states a removal, a change, or anything that happened, only the standing fact. I also swept every line of prose I added this round for the same shape and rewrote three ("as the confinement was", "was never fetchable", "the one that was classified") into standing present tense before committing.

### 4. Close the fail-open in the Action's network gate

`action.yml` now reads `if [ "$SKILLSAW_NO_NETWORK_INPUT" != "false" ]`, so `'True'`, `'yes'`, `1` and any typo keep the network off. Left `no-custom-rules` alone — noted as optional, and it has no fresh threat entry behind it. `docs/ci.md` and the rule doc now say the literal `false` is what opts a job back in.

### 5. Keep a malformed `Location` header inconclusive

`super().http_error_302(...)` is wrapped in `try/except ValueError`, re-raised as `urllib.error.HTTPError(req.full_url, code, msg, headers, body)` — the same shape a refused hop produces, carrying the `_EmptyBody` wrapper so nothing gets drained. `ValueError` stays out of `_NETWORK_ERRORS`.

New route `"/redirect-malformed": (302, "http://[::1/x")` and `test_a_malformed_location_header_is_inconclusive`, which asserts the premise (`pytest.raises(ValueError)` on `urlparse("http://[::1/x")`), that only `/redirect-malformed` was requested, and that the probe comes back `checked=True` with a status that is not 404/410. Confirmed non-vacuous against the base class:

```
guarded   -> HTTPError 302
unguarded -> ValueError: Invalid IPv6 URL
```

### Hermeticity re-verified

`tests/test_external_links.py` is 196 tests, all green, and the whole suite still runs offline: `conftest.py`'s socket guard is untouched, `test_default_run_cannot_open_a_socket` still holds, and `fix` has moved from "gateable" to "structurally offline".

docs/rules/content-broken-external-reference.md
stbenjam, line 1: This file is absolutely insanely verbose

stbenjam:
## Round-4 response

All five Required Actions are landed in `94a6daf`, each verified non-vacuously by reverting the fix and confirming the targeted tests fail. `make test` 4993 passed, `make lint` clean, `make update` regenerated `docs/cli.md` and the site rule-doc copy. Self-lint stays A+ / 0 errors / 0 warnings.

**Also in this push, and not an aside:** `d6ef112` implements the maintainer's stated requirement — *"The broken reference external thing needs to be exposed in the onboarding skill. Needs to be possible to be used in the action and you should offer an ability to schedule a recurring task to run this occasionally."* The panel filed those as Optional Follow-ups (Ecosystem 1–2); they are not optional. Details at the bottom, so round 5 does not read that commit as scope creep.

---

### 1. The two remaining holes in `_is_public_host`, and the docs that claimed otherwise

**(a) Leftover `%` is refused.** `_canonical_host` now returns `None` when a `%` survives `unquote(...).encode("idna").decode("ascii")`. Reproduced the arbiter's table before fixing, and the single-encoded full-width `％` row is the one that mattered most — NFKC inside the codec folds it to `%`, so no double encoding is needed.

**(b) `_is_public_host` canonicalizes first, then strips.** Userinfo, port and brackets come off the *canonical* string, so a delimiter canonicalization introduces is parsed rather than swallowed into the "it's a name → public" branch. This is defence-in-depth once (a) is in — `_request_url` refuses these upstream — so I verified it independently by calling `_is_public_host` directly:

```
169.254.169.254%3A80    _is_public_host=False
127.0.0.1%3a8080        _is_public_host=False
localhost%3A80          _is_public_host=False
10.0.0.5%3A22           _is_public_host=False
```

**(c) `not address.is_global` added alongside the existing six**, not replacing them — they catch multicast and NAT64 where `is_global` is True. Layered, so drift in `is_global` can only tighten.

**(d) Userinfo check moved after `_canonical_netloc()`; control characters refused** in `_canonical_host` via `_FORBIDDEN_HOST_CHARS` (the `[\x00-\x20\x7f]` set `http.client._validate_host` uses). The confinement now reaches the NUL verdict on its own reasoning rather than being saved by a downstream CVE fix.

**Tests.** The inline refusal list became `_REFUSED` and now carries all eight cases from the verdict. `test_non_canonical_ipv4_redirect_hops_are_refused` is renamed `test_refused_hosts_are_refused_as_redirect_hops_too` and parametrizes over `_REFUSED + _ALTERNATE_SPELLINGS` — an origin picks the `Location` header, so a hop is the *more* hostile path, not the less. `_on_the_wire()` now models **both** decodes and returns `(authored, emitted, wire)`, deriving the wire host from `http.client.HTTPConnection` so it is an independent statement of the transport's spelling. Two dedicated tests assert the premise from the stdlib before asserting the refusal, so neither can rot into passing vacuously:

```
1a REVERTED  -> 5 failed (second_percent_decode + refusal rows)
1c REVERTED  -> 2 failed (100.64.0.1, 100.100.100.100)
1d-i REVERTED -> 2 failed (127.0.0.1%00.evil.com + control-char test)
1d-ii REVERTED -> 2 failed (user%40…, 127.0.0.1%40…)
```

**Docs.** T18 and the rule doc enumerations now list unspecified and "anything `ipaddress` does not call global", with the CGNAT reasoning spelled out, plus the post-encoding refusals. The `and the reported URL` clause is struck from T18, the rule doc and the `test_the_operator_can_still_reach_them` docstring — Writer 1 is right that reporting the authored href is the correct behaviour and only the sentence was wrong. The rule doc now says so positively: the report names the string you have to find in the file, and it differs from the wire routinely.

### 2. `skillsaw fix --rule <network rule>`

Took the "tailor the message" branch. `_apply_network_gate` raises `EmptyNetworkRuleSetError` (a `ValueError` subclass, so every existing handler is unaffected) carrying `rule_ids`; `_fix.py` catches it and answers with advice that works:

```
Error: fix never runs network rules: content-broken-external-reference.
Run skillsaw lint --rule content-broken-external-reference to check them.
```

Pinned by QA's `test_fix_with_only_a_network_rule_named_explains_itself`, asserting `returncode == 1`, `"drop --no-network" not in stderr` and `"skillsaw lint --rule <id>" in stderr`.

On `--allow-private-hosts`: I kept `_add_network_flags(fix_parser)` and gave `fix` its own help note rather than dropping the flags. Dropping them would make `skillsaw fix --allow-private-hosts` an argparse error, which would quietly gut `test_fix_never_probes_whatever_the_flags_say` — its whole premise is that the flags say one thing and `fix` ignores them. `docs/cli.md` is regenerated and now reads "Accepted on fix for argv compatibility only: fix never runs network rules, so this has no effect there". Parser, docs and behaviour now agree. I also took QA 2's `assert result.returncode == 0, result.stderr` at that site, since it is what makes the negative leg load-bearing.

### 3. Review-history comment

`tests/test_external_links.py` → `# The cloud metadata endpoint — the canonical SSRF target (T18).`

### 4. `ValueError` guard confined to the parse

`http_error_302` now validates `headers.get("location") or headers.get("uri")` with `urlparse()` in a small `try/except` at the top, raises the `HTTPError` there, and delegates to `super()` **unguarded**. Mirrors the base class, which does `urlparse(newurl)` as its first act. The existing malformed-`Location` test passes unchanged.

### 5. `_is_ignored` bounded

`_MAX_URL_LENGTH = 2048` (refused in `_request_url`) and `_MAX_IGNORE_PATTERN_LENGTH = 256` (pattern skipped), both in the module's existing `_MAX_*` style, with the comment recording why SIGALRM cannot transplant — `_accepts_hop` runs on worker threads. `test_both_operands_of_an_ignore_glob_are_length_bounded` covers both halves and deliberately uses a benign over-long pattern for the skip leg, since a pathological one would hang rather than fail on revert.

### Zero network by default, re-verified

Instrumented `socket.socket.connect` and ran the real CLI:

```
lint, DEFAULT config (rule off)              sockets=[]
lint, rule ENABLED, no flags                 sockets=[('172.66.147.243', 443)]   # example.com only
fix,  rule ENABLED, no flags                 sockets=[]
fix,  rule ENABLED, --allow-private-hosts    sockets=[]
lint, rule ENABLED, --no-network             sockets=[]
```

The one connection is the public host; `169.254.169.254` and `100.64.0.1` were both refused in the same run — the new `is_global` disqualifier working end to end, not just in a unit test.

---

### `d6ef112` — the maintainer's discoverability requirement

Included in this push and **intentional**, addressing stbenjam's requirement directly rather than the panel's optional framing. Verified against the diffs, all three asks satisfied:

1. **Exposed in onboarding.** `skills/skillsaw-onboard/SKILL.md` Step 7 gains a third sub-branch in the same "Then ask if they …" shape as the existing PR-review offer directly above it, writing `.github/workflows/link-check.yml` only on yes, and adding it to the Step 9 commit reminder. Mirrored to `.claude/` and `.agents/`.
2. **Usable from the Action.** New `rule` input → `SKILLSAW_RULE` → validated kebab-case ids appended as `--rule`. `ARGS` is word-split unquoted, so the guard refuses anything outside `[a-z0-9-]` under `LC_ALL=C`; `shell: bash` is confirmed, so the `<<<` herestring and `${VAR//,/$'\n'}` are sound.
3. **Recurring schedule offered.** The offer is for a weekly `schedule` cron plus `workflow_dispatch`, as its own workflow so a slow origin never blocks a merge, pointing at `docs/ci.md#scheduled-external-link-checking` — anchor confirmed against the `## Scheduled external link checking` heading.

`tests/test_workflow_hardening.py` pins all of it: line/comma splitting, hostile-id rejection across locales, that `rule` alone does **not** grant network access (selection is not permission), and that the `docs/ci.md` recipe and all three skill copies agree. That last one also closes QA 4's request for `no-network` wiring guards in that file.

Ecosystem 2's sharper point is resolved as a side effect: `no-network: false` now has a sanctioned use, because the Action can express `--rule`, so the scheduled recipe is `uses: stbenjam/skillsaw@…` again rather than `pipx install skillsaw`.

---

Nothing outside these was changed — the remaining Optional Follow-ups are untouched and still open.

stbenjam:
## Round-5 response — all five required actions landed (`54ddee5`)

Two sub-clauses deliberately trimmed; called out below. Also a size pass on the diff, per the maintainer.

---

**1. Guard the rebuild's re-parse.** Both parses now live under one `try` in `_request_url`, which returns `(url, netloc)` — `_admit` takes the netloc it was handed instead of re-splitting, so that is one guard and one fewer `urlsplit` per URL and per hop, as asked. Handing back the *parsed* netloc rather than `_canonical_netloc`'s output is also strictly more faithful: a decode that introduces `/`, `?` or `#` shortens the authority `urlsplit` (and `Request._parse`) sees, and the classifier now follows it.

`http://example.com%5B/x`, `http://example.com%5D/x` and `http://%EF%BC%BB127.0.0.1%EF%BC%BD/x` are in `_REFUSED`, so they flow into `test_refused_hosts_are_refused_as_redirect_hops_too` too. Non-vacuity checked by moving the re-parse back outside the `try`: 1 failure.

**2. Validate the port half.** `_is_port` — empty, or a colon and ASCII digits — gates both branches. Not `str.isdigit()`: that is true for the full-width digits nameprep folds to ASCII ones, which is the substitution this rule refuses everywhere else. The four requested rows are in `_REFUSED`, and `test_a_port_that_decodes_into_something_else_is_refused` asserts them under `allow_private_hosts=True` with a stdlib liveness premise (`unquote(authority) == the decoded string`), the way the second-decode test does. It carries a fifth row, `http://[2606:4700::1]:8%300/x` — without it the *bracketed* branch's check has no test that binds, since every other bracketed row is refused on its host.

**Trimmed:** the optional "make `_is_public_host` return `False` on a surviving `@`". The port validation already makes `rsplit("@", 1)` unreachable rather than wrong, which was the stated goal.

**3. Lowercase the canonical host.** `_canonical_host` returns `canonical.lower()`. The docstring records why it has to be done here (the ASCII fast path case-folds nothing) rather than restating that DNS is case-insensitive. `test_ignore_sees_the_url_the_transport_will_request` gained `https://Intranet.Example.COM/wiki`, and `test_request_url` gained a case-mixed row; the pre-existing `HTTPS://Example.com/a` row now expects a lowercased host. Reverting the `.lower()` fails 1.

**4. The three documentation claims.**
- **IPv4-mapped** is out of the refusal list in T18 and the rule doc, replaced with the unwrap clause: `::ffff:127.0.0.1` is refused as loopback, `::ffff:8.8.8.8` is treated like any other public address.
- **The `_MAX_*` bounds** are documented in the rule doc's `ignore` section (a pattern over 256 characters is skipped and the URLs it would have matched are probed; a URL over 2,048 characters is never requested; neither is reported) and in T18's clamping inventory.
- **The ReDoS rationale** is gone from the module comment and the test docstring. Both now say what the bounds do — keep both `fnmatch` operands small, and for the URL also bound the report and the de-duplication key it becomes — and note that `fnmatch.translate` has emitted the linear emulation since 3.9 (bpo-40480). The bounds themselves are unchanged.

The port validation is a new refusal, so it got one clause in the rule doc and one in T18 rather than shipping undocumented.

**5. Onboarding block.** The directive now sits immediately above the line carrying the `` `.github/workflows/link-check.yml` `` code span. The workflow is named `link-check` in the skill and in `docs/ci.md` (I used the recipe's exact spelling rather than `Link check`, so the two cannot drift), with the reason inline; the skill tells the agent to reuse the SHAs it already looked up. `make update` remirrored all three copies, and `test_scheduled_link_check_recipe_matches_the_onboard_skill` now slices the skill to the new section — `strict: true` and `name: link-check` both bind there, on both sources.

One correction, since I would rather report it than quietly "fix" it: **that directive was not producing an INFO in this repo, in either position.** `content-unlinked-internal-reference`'s default `patterns` are `["./**/*.*", "references/**/*.md"]`, and `.github/workflows/link-check.yml` matches neither — I verified by deleting the directive and re-linting the repo (0 violations either way). The placement fix is still right as authoring, and it is kept.

**The self-lint defect is real, though — it is two lines away.** `.apm/instructions/development.instructions.md:108-109`, in the `### Network-touching rules` block this PR adds, reference `src/skillsaw/rules/builtin/content/broken_external_reference.py` and its rule doc by full path, and *those* do match. Branch: 5 `content-unlinked-internal-reference` INFOs; `main`: 3. The block now names the files without full paths, and the branch is back to 3 — no new self-lint finding anywhere in the PR.

---

### Size

The maintainer's separate instruction this round was that ~4,500 additions is too many, so I did a consolidation pass alongside the fixes. **4,488 → 4,405**, and that is net of roughly 120 lines of *new* required-action code, tests and docs — so the consolidation removed about 200. No test case was dropped; near-identical test functions became denser tables:

- three `_methods_by_path` assertions over three separate full lint runs → one dict, one run (and the `/head-404-get-200` / `/head-410-get-200` / `/missing` method rows fold in with it)
- `--no-network` and `SKILLSAW_NO_NETWORK` × three subcommands → one test parametrized on both axes; the redundant `test_env_var_is_honoured` (identical to the `"1"` row of the typo table) is gone, and the on/off tables merged into one
- three clamp tests with identical `_probe_all` capture boilerplate → one `_as_applied` helper and one 17-row table
- three "what activates a disabled-by-default rule" tests → one table
- the two subprocess announcement tests, the two hop-refusal tests, the two `_request` exception tests, the four `TestMetadata` assertions, the two budget-notice tests, the two out-of-scope markers, the hop-cap pair, and the handler's four split-brain routes → merged
- rule doc: the duplicated "any setting enables it" paragraph and the duplicated `--allow-private-hosts` forward reference removed, prose tightened (303 → 289, and it mirrors, so twice that off the diff)

What is left is not duplication. The network-gate wiring across the CLI, the confinement logic, and the reasoning comments four rounds of review were verified against are the bulk of it, and I did not touch them.

### Validation

`make test` 4,996 passed · `make lint` clean · `make update` regenerated and committed · repo self-lint A+, 0 errors, 0 warnings, and no new INFO vs `main` · `openshift-eng/ai-helpers` exit 0 (its one ERROR is `plugins-doc-up-to-date` failing to find `skillsaw` on PATH — identical on `main`) · zero-network-by-default re-verified: the whole suite runs under the `conftest.py` socket tripwire, and the hermeticity and gate tests (46) pass · `origin/main` merged before push.

PR #522, opened 2026-08-28
Add claude-md-agents-import: recommend @AGENTS.md as the whole CLAUDE.md

stbenjam:
## Summary

Adds `claude-md-agents-import` (INFO, `enabled: auto`, `since 0.20.0`): when a
`CLAUDE.md` and an `AGENTS.md` share a directory, recommend that `CLAUDE.md`
be exactly the one line `@AGENTS.md`.

This is the #2 ecosystem pain point from #513: Claude Code reads `CLAUDE.md`,
almost everything else reads `AGENTS.md`, and teams that keep both maintain
two copies that drift. Claude Code's `@path` import syntax removes the
duplication instead of policing it — one source of truth, every assistant
reading the same words.

**Tolerance, stated once:** blank lines and HTML comments never count as
content; everything else does (headings, prose, code fences, a second
import). So a licence banner or an inline suppression directive above the
import keeps the file "import-only". Comment spans come from the markdown
AST (`prose_lines()` blanks them column-for-column, preserving the line
count) — no hand-rolled regex.

**Autofix:** `SUGGEST`, and only when the two bodies are identical after
per-line trailing-whitespace normalization, where the import provably loses
nothing. Everything non-identical is report-only (`fixable=False`) — skillsaw
does not decide which of your sentences survive. The fix deliberately changes
the file's line count: it is a file-restructuring fix like the frontmatter
ones, not an in-place splice, and the tests say so instead of asserting line
preservation.

**Config** (both fell out naturally, so both shipped):

| option | default | why |
|---|---|---|
| `allow-extra` | `false` | accept the import *plus* Claude-specific sections; only a CLAUDE.md with no import at all is then reported |
| `ignore-generated` | `true` | a CLAUDE.md compiled by APM or similar cannot be hand-replaced — the next build overwrites it. Mirrors `content-instruction-drift`'s option of the same name |

Note `_is_content_rule()` in the linter keys compiled-copy suppression on the
`content-` prefix, so this rule does not inherit it; `ignore-generated` is its
own guard rather than a duplicate of one.

## The drift-rule interplay

`content-instruction-drift` and this rule are complementary, and the PR pins
that with tests rather than asserting it in prose:

- **Drift is the detector; this rule is the cure.** Drift compares sections
  across instruction files and reports near-duplicates that grew apart — a
  real finding *after* the copies exist. This rule fires earlier, on the
  existence of the duplicated pair, and recommends the structure under which
  divergence is impossible.
- **No teaching was required.** The recommended end state — an import-only
  `CLAUDE.md` — carries no headings, so `_extract_sections()` yields nothing
  and it contributes no comparable sections. Drift is structurally silent
  there. I inspected the rule expecting to have to teach it that an
  import-only file is resolved; it already is, and
  `test_instruction_drift_stays_silent_on_the_end_state` pins that so a
  future change to `_extract_sections` cannot quietly break it.
- **Identical copies:** drift skips them by design ("identical copies are
  intentional sync"), which leaves an exactly-duplicated pair reported by
  *nobody* today. That is the gap this rule closes — and it is the case where
  the fix is safe. Pinned in both the unit and integration suites.
- **Diverged copies:** both rules fire, and that is correct. Drift names the
  section that disagrees; this rule recommends the layout that makes the
  disagreement impossible. `test_diverged_pair_reports_both_rules` asserts it.
- **Opt-out is coherent:** a team that would rather keep the copies and be
  told when they diverge disables this rule and keeps drift. Documented.

`instruction-imports-valid` needed no change either: `@AGENTS.md` resolves to
a sibling file, so it passes. This rule never emits an import diagnostic of
its own — a broken `@AGENTS.mdx` is reported by imports-valid alone, and reads
here as ordinary content
(`test_a_broken_import_is_reported_only_by_imports_valid`).

## Deviations from the spec

1. **"Reuse the pairing logic `content-instruction-drift` uses."** There is
   none to reuse: drift compares every instruction block pairwise across the
   repository and never pairs per directory. `_pairs()` groups
   `find(ClaudeMdBlock)` / `find(AgentsMdBlock)` on their resolved parent
   directory instead, with the reason in the docstring.
2. **"Root or any shared subdirectory."** Instruction-file discovery
   (`scan_repository`) is root-scoped for `AGENTS.md`/`CLAUDE.md` — unlike
   `.cursor/` directories, which are walked. A subpackage pair is invisible to
   *every* instruction rule today, this one included. The rule pairs on the
   resolved directory rather than assuming the root, so subpackage pairs are
   reported unchanged the day discovery starts walking;
   `test_pairing_follows_root_scoped_instruction_discovery` pins today's scope
   so the difference is visible rather than surprising.
3. **Symlinks — found against ai-helpers, not in the spec.**
   `openshift-eng/ai-helpers` ships `CLAUDE.md` as a *symlink* to `AGENTS.md`:
   one file under two names, the other honest answer, with no second copy to
   drift. The lint tree already dedupes on resolved path so the rule never saw
   it, but relying on that is fragile — `_pairs()` now states the intent
   directly, with a regression test and a note in the rule doc (including why
   the import is still the recommendation: symlinks are not portable to every
   checkout).
4. **Empty `CLAUDE.md` is skipped**, not reported: `instruction-file-valid`
   already owns that finding and there is no content to move.
5. **`config_schema`** ships both `allow-extra` and `ignore-generated` rather
   than deferring; see the table above.
6. **Rule id.** Implemented as specified. For the record, the repo's naming
   convention does argue both ways: `claude-*` is the Claude Code *format*
   family (plugin manifests, command/agent frontmatter, settings), while the
   `instruction-*` family is exactly "rules over AGENTS.md/CLAUDE.md/
   GEMINI.md/QWEN.md" — which is what this rule is. `instruction-agents-import`
   would sit more naturally beside `instruction-imports-valid`, and the rule
   lives in that package (`rules/builtin/instructions/agents_import.py`) and
   that docs group. I kept `claude-md-agents-import` per the spec because it
   names the file it lints and reads well in output; happy to rename before
   merge — nothing has shipped under the id yet, so no alias would be needed.

## Test plan

`make test`: **4706 passed**, 0 failed (4695 before this branch).

New coverage:

- `tests/test_instruction_file_rules.py` — `TestClaudeMdAgentsImportRule`
  (metadata; CLAUDE-only / AGENTS-only / already-the-import / `@./AGENTS.md`;
  comment + blank-line tolerance; empty file; generated marker both ways;
  symlink; root-scoped discovery; duplicated fires and is fixable;
  trailing-whitespace-only difference still fixable; diverged fires not
  fixable; reported line is the first non-import line; code fence counts as
  content; `allow-extra` both directions; fix replaces / declines / is
  idempotent).
- `TestClaudeMdAgentsImportInteractions` — the three interaction pins above,
  plus `test_no_builtin_rule_fires_on_the_import_only_layout`, which runs
  **every** builtin rule against the import-only layout and asserts zero
  violations. That is stronger than the default rule set and is how the
  "minimal-file content rules" question is answered: no rule has a
  minimum-content expectation that trips on the one-liner, so nothing needed
  resolving.
- `tests/test_integration.py` — `TestClaudeMdAgentsImport`, nine end-to-end
  tests over four fixtures under
  `tests/fixtures/instructions/agents-import/`: `duplicated-pair`,
  `diverged-pair`, `import-only`, `import-plus-extras`. Covers the
  `--suggest` gate (plain `skillsaw fix` changes nothing), convergence to the
  import-only file, a clean re-lint, idempotency, the `allow-extra` config
  path, and the fix declining the diverged pair.
- `duplicated-pair` joins `BROKEN_FIXTURES` and `import-only` joins
  `CLEAN_FIXTURES`, so `test_every_rule_fires_somewhere` and
  `test_all_clean_fixtures_pass` cover the rule.

Docs: `src/skillsaw/rules/docs/claude-md-agents-import.md` positions this as
the CLAUDE.md↔AGENTS.md answer (#513 rec #6) — the why, the `@AGENTS.md` end
state, the drift interplay, and every escape hatch (`allow-extra`,
`enabled: false`, inline suppression, generated files, symlinks).

## Validation

- `make test` — 4706 passed.
- `make lint` — clean (black, 297 files).
- `make generate-example generate-docs generate-site-content` — regenerated;
  rule count 78 → 79, Instruction Files group 2 → 3 rules.
  `make verify-apm` — OK (no `.apm/` sources touched). `make badge` — no
  change. `skillsaw lint .` (self-lint) — 0 errors, 0 warnings, A+.
- **Benchmark** (medium, baseline saved on `origin/main` @ c70a542, compared
  on this branch): **no regressions detected**. `rules_run` min 597.7ms →
  598.4ms; total min 771.9ms → 787.3ms (within noise, far under the 25% gate).
  The new rule does not appear in the top-15 slowest. Synthetic-repo
  violations went 150 → 151 — the benchmark corpus has a duplicated pair, so
  the rule does fire there.
- **openshift-eng/ai-helpers** — `skillsaw lint .` **exits 0**, unchanged
  (0 errors, 0 warnings, A+).
- **Force-enabled against ai-helpers**
  (`skillsaw lint --no-baseline --rule claude-md-agents-import .`):
  **0 findings, exit 0.** ai-helpers does have a root `CLAUDE.md` *and*
  `AGENTS.md`, and the CLAUDE.md content is byte-identical — but it is a
  symlink, so there is genuinely one file and nothing to report. That is the
  deviation-3 case above; it is the only CLAUDE.md/AGENTS.md pair in the
  repository. Running the rule alongside `instruction-imports-valid` and
  `content-instruction-drift` on the same checkout also yields 0.

## Note on `since`

`since = "0.20.0"`, which is the current `pyproject.toml` version and is
unreleased on PyPI. `test_every_rule_since_is_not_in_the_future` passes today.
If 0.20.0 ships before this merges, the `since` needs bumping to the next
unreleased version or the rule is silently disabled for every version-pinning
user.

## Parallel work

Three sibling branches (OpenCode support, Roo support, external link
checking) are in flight. Expect conflicts in the generated docs and rule
counts at merge time; this branch stands alone against `origin/main`
(merged through c70a542).

Assisted-by: Claude:claude-opus-5

https://claude.ai/code/session_01J1Qt3WNQCWATbFrjxLQZPL

## Summary by CodeRabbit

- **New Features**
  - Added a validation rule for adjacent `CLAUDE.md` and `AGENTS.md` files.
  - Detects duplicated or diverging instructions and recommends using `@AGENTS.md` in `CLAUDE.md`.
  - Supports generated-file handling, symlink scenarios, extra-content configuration, suppressions, and informational reporting.
  - Added a suggested autofix for exact duplicate files.

- **Documentation**
  - Added comprehensive rule guidance, examples, configuration details, and troubleshooting information.
  - Updated the rules index and Instruction Files documentation to include the new rule.


stbenjam:
@coderabbitai review

docs/rules/claude-md-agents-import.md
stbenjam, line 1: This is way too wordy

docs/rules/instruction-files.md
stbenjam, line 1: Don't mention the other rule here

stbenjam (commented):
Some minor nits but lgtm otherwise, please fix these

docs/rules/claude-md-agents-import.md
stbenjam: Fixed in 494c0c6. Cut from 148 lines to 63 — it was the longest rule doc of all 78 (median is ~40). Kept the why, one bad/good pair, the tolerance in two sentences, how to fix, and the escape hatches; the drift interplay is now three sentences instead of a section.

docs/rules/instruction-files.md
stbenjam: Fixed in 494c0c6. Reverted the group blurb in both doc generators, so this page prose is now byte-identical to main — only the generated table row is added. (The rule id still has to be listed in RULE_GROUPS or generate-site-content.py fails on an uncategorized rule, but that adds no prose.)

src/skillsaw/rules/builtin/instructions/agents_import.py
```diff
@@ -0,0 +1,301 @@
...
+            body = claude.read_body(strip_code_blocks=False)
+            if not body or not body.strip():
+                # An empty CLAUDE.md is ``instruction-file-valid``'s finding,
+                # and there is no content to move anywhere.
+                continue
+            if ignore_generated and has_generated_marker(body):
+                continue
+
+            shape = self._shape(claude, agents.resolved_path)
+            if not shape.extra_lines:
```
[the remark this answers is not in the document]
stbenjam: Weird edge case, no

src/skillsaw/rules/builtin/instructions/agents_import.py
```diff
@@ -0,0 +1,301 @@
...
+
+    @staticmethod
+    def _normalized(body: Optional[str]) -> List[str]:
+        """Body lines with trailing whitespace stripped — the identity test.
+
+        Anything finer than trailing whitespace is byte-meaningful: if the
+        two files differ at all, replacing one with an import is a content
+        decision a linter must not make on its own.
+        """
+        return [line.rstrip() for line in (body or "").splitlines()]
```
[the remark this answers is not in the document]
stbenjam: Sounds fine to me

src/skillsaw/rules/builtin/instructions/agents_import.py
```diff
@@ -0,0 +1,301 @@
...
+        base = block.path.parent
+        shape = _BodyShape()
+        for index, line in enumerate(raw):
+            if not line.strip():
+                continue
+            body_line = index + 1
+            blanked = index < len(prose) and not prose[index].strip()
+            if blanked and body_line in comment_lines:
+                continue
+            if self._is_sibling_import(line, base, agents_resolved):
```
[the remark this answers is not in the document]
stbenjam: Another weird edge case

PR #529, opened 2026-08-28
Cut the test suite runtime in half

stbenjam:
## What was slow

Profiling showed the suite made **745 CLI subprocess spawns costing 174s of wall time**, with no hot spot — a long tail of ~260ms spawns, roughly **60% of which was interpreter startup and importing skillsaw** before a single file got read. `tests/test_integration.py` alone accounted for 536 of them.

## Results

Measured on an 8-core darwin box, `-n auto`:

| | before | after | |
|---|---|---|---|
| `make test` | 31.1s / 263s CPU | **15.0s / 124s CPU** | **2.1x** |
| `make test-coverage` | 50.6s / 398s CPU | **20.4s / 143s CPU** | **2.5x** |

Coverage is unchanged: 94%, 1008 missed lines, same per-file numbers across the `cli/` package.

## Three changes

**1. Import `jsonschema` lazily.** `agent_plugins/_helpers` imported it and compiled four `Draft202012Validator`s at module load — ~20ms on *every* run, including the overwhelming majority that lint a repository with no `plugin.json` at all. Schemas, validators and manifest fields now resolve through cached accessors. This speeds up the shipped CLI for users too, not just the tests.

**2. Measure coverage with the `sys.monitoring` backend** (`core = "sysmon"`). It traces far more cheaply than the `settrace` core, which matters while the suite still spawns instrumented subprocesses that pay the tracing cost from interpreter start. Set in `pyproject.toml` rather than the Makefile so subprocesses and a bare `pytest --cov` inherit it. Coverage warns and falls back on an interpreter that can't use it, so contributors on 3.9–3.11 are unaffected.

**3. Drive the integration CLI in-process.** `tests/cli_runner.run_cli` calls `skillsaw.cli.main()` with argv, stdout, stderr, env and cwd swapped and returns a `CompletedProcess` stand-in. Only the four helpers in `test_integration.py` were repointed at it — a 20-line diff.

## What deliberately keeps `subprocess.run`

A prototype that converted *everything* broke 22 tests, and they were right to break:

- **`--no-custom-rules` tests** prove a rule file on disk did or did not get imported. In-process they would execute an arbitrary rule inside the test worker, and `sys.modules` could mask a repeat import and make the negative case pass vacuously.
- **Lone-surrogate regressions** assert that rendering a report raises no `UnicodeEncodeError`. A `StringIO` never encodes, so in-process they would pass vacuously.
- **The colour cascade** depends on the real stream.
- **`test_integration_plugins.py`** installs plugins and needs a fresh interpreter to discover their entry points — untouched.

That reasoning is written into `.apm/instructions/testing.instructions.md` (propagated by `make update`) so the next person adding a test knows which side to pick.

## Risks

- **Sharing one interpreter is a new hazard.** `run_cli` restores the two globals that outlive a call: root logging handlers (`basicConfig` is a no-op once handlers exist) and the `skillsaw.utils` file cache (only invalidated inside the autofix loop, because a real run starts with it empty). Verified by running the **whole suite serially in a single process** — 4706 passed. A future helper that leaks global state will break later tests, hence the documentation.
- **Converted calls lose `subprocess.run`'s `timeout=60`.** A rule that loops forever now hangs the suite rather than failing one test. The CI job timeout still catches it.

## Checklist

- [x] Up to date with `upstream/main` (21cc75a)
- [x] `make test` — 4706 passed (also passes fully serially)
- [x] `make lint`
- [x] `make update` — self-lint clean
- [x] `openshift-eng/ai-helpers` exits 0

https://claude.ai/code/session_01Kf41faZ6bv3qPudU3XAPGz


stbenjam:
Wild

Issue #532, opened 2026-08-28
[UX] Audit default-rule first-run noise for 0.20.0


not-stbenjam:
## Scope

This is a narrow 0.20.0 release-readiness audit of **first-run default-rule noise**. It is not a general ecosystem roadmap and does not implement any policy change. Every recommendation below requires maintainer approval.

I consulted #513 and its comments (there are currently no comments). That issue was used as historical context only; this report intentionally does not repeat or reopen its ecosystem/tool-support recommendations.

## Methodology

- skillsaw: `0.20.0` from latest `origin/main`, commit [`f69147e`](https://github.com/stbenjam/skillsaw/commit/f69147ed1bb787fcb6b3c5beea2ba9eaf7669cd0)
- Audit date: 2026-08-28
- Six public repositories, shallow-cloned at the exact commits below.
- Standard first-run command (JSON also retained for analysis):

  ```text
  .venv/bin/skillsaw lint <repo> --no-custom-rules --no-network --no-progress --format json
  ```

- A second `-v` JSON pass exposed info-level findings for rule-distribution and grade analysis. The ordinary run hides those individual findings but still prints their count and includes them in the grade.
- All repository content was treated as untrusted data, not instructions.
- Five repositories had neither `.skillsaw.yaml` nor `.skillsaw-baseline.json`, so their run is a genuinely unpinned first run.
- `auth0/agent-skills` was the configuration-control repository. I captured its existing config run, moved `.skillsaw.yaml` out of the checkout, confirmed the no-config run, then restored it. No baseline existed. Because `AGENTS.md` links to `.skillsaw.yaml`, literal removal creates one artificial broken-link finding; I also ran with a minimal external `version: "0.20.0"` config while leaving the linked file in place to isolate the default-policy delta.
- I manually inspected representative source context for every high-volume warning/error rule, all five secret errors, and samples of the highest-volume info rules. Counts alone were not used to classify noise.

## Corpus and first-run totals

Popularity/activity signals are GitHub API values captured 2026-08-28.

| Repository | Commit | Signal | Config state | Errors | Warnings | Info (hidden) | Grade | Text output |
|---|---|---:|---|---:|---:|---:|---:|---:|
[...]

PR #535, opened 2026-08-28
Recognize direct description routing phrases

not-stbenjam:
## Summary

- recognize clear sentence-level routing clauses beginning with Use for and Use it when
- preserve sentence-boundary guards against explanatory prose false positives
- keep missing, empty, and name-restatement diagnostics unchanged
- add realistic regression fixtures and update rule documentation

This implements the focused signal recommendation from #532. Findings on the sampled corpora drop from 8 to 6 for anthropics/skills and 242 to 226 for github/awesome-copilot.

## Validation

- make test: 5,130 passed
- make lint
- make update and self-lint
- focused routing and marketplace tests: 101 passed
- medium benchmark: no regression
- latest openshift-eng/ai-helpers: exit 0

### DIFF
--- a/docs/rules/content-description-routing.md
+++ b/docs/rules/content-description-routing.md
@@ -26,9 +26,9 @@ to them, so this rule skips them by default. Set `check-user-only-skills: true`
 to check their descriptions normally. Only the YAML boolean `true` opts a skill
 out; an absent field, `false`, strings, and numbers remain checked.
 
-Natural selection clauses count as trigger phrasing, including "Use this skill
-for ...", "Invoke this skill whenever ...", "This skill should be used before
-...", and "Use only when ...".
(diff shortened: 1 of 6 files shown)


stbenjam:
I believe I've seen "use proactively for..."

I wonder if "Use (small distance) when/for" is useful?

not-stbenjam:
The current bounded matcher already covers this family: `Use proactively for ...` and `Use this skill proactively when ...` both match (as do `Use it when ...` and direct `Use for ...`). I would keep the gap vocabulary-bounded rather than allow an arbitrary small word distance: otherwise instructional prose such as `Use standard terminology when ...` can satisfy the routing check even though it does not say when to select the skill. So I think the current scope captures the useful forms without opening that false-negative class.

PR #537, opened 2026-08-28
Collapse repeated MCP tool names in text output

not-stbenjam:
## Summary

- collapse three or more content-mcp-tool-name findings sharing a file and server prefix into one terminal row
- preserve every underlying occurrence for fingerprints, baselines, fixes, counts, exit status, grade, and structured reports
- retain actionable first-location context plus up to three distinct sample lines
- document the presentation-only behavior and cover threshold, metadata isolation, malformed discriminators, baselines, and fix convergence

## Stack

This PR is intentionally based on #536 because it reuses that PR's generic text-collapse infrastructure. Its diff is MCP-only. After #536 merges, this PR should be retargeted to main and updated from latest main before merge.

## Validation

- make test: 5139 passed
- make lint
- make update and self-lint
- latest openshift-eng/ai-helpers: exit 0
- focused formatter and MCP rule tests: 126 passed
- focused integration tests: passed

### DIFF
--- a/docs/rules/content-mcp-tool-name.md
+++ b/docs/rules/content-mcp-tool-name.md
@@ -109,6 +109,10 @@ A violation in a body decoded out of a non-markdown host — a JSON hook
 prompt, a folded (`>`) YAML scalar — is reported without an automatic fix;
 apply the same rewrites by hand.
 
+When three or more findings in one file share a server prefix, terminal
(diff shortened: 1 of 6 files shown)


not-stbenjam:
Folded into #536 so both text-collapse behaviors use and review the same shared formatter mechanism in one PR.

PR #543, opened 2026-08-29
Add a dedicated scheduled link-check Action

not-stbenjam:
## Summary

- ship `stbenjam/skillsaw/link-check` with the safe external-link settings built in
- create, update, and reopen one owned GitHub issue for confirmed 404/410 findings without auto-closing on inconclusive silence
- switch `skillsaw-onboard` to progressive disclosure with one reference per step and have its CI step install the packaged Action
- document the thin scheduled workflow, permissions, inputs, and `GITHUB_TOKEN` event behavior
- add reporter lifecycle, Action hardening, documentation consistency, and skill-layout regression tests

## Validation

- `make test` — 5,152 passed
- `make lint`
- `make update` — repository self-lint A+, 0 errors, 0 warnings
- `make verify-apm`
- focused Action/reporter/release tests — 29 passed
- fresh `openshift-eng/ai-helpers` clone — exit 0, 0 errors, 0 warnings

PR #548, opened 2026-08-29
Remove unreleased external link checking

not-stbenjam:
## Summary

- remove the unreleased external-reference rule and every network-specific CLI, API, Action, CI, security, documentation, fixture, and generated surface it introduced
- restore the contract that skillsaw itself makes no network calls
- keep external URL checking outside skillsaw by recommending a scheduled Lychee Action workflow
- split skillsaw-onboard into a small router plus per-step references for progressive disclosure

This supersedes #543.

## Validation

- make test: 4,932 passed
- make lint
- make update
- make verify-apm
- skillsaw self-lint: A+, 0 errors, 0 warnings
- skillsaw-onboard lint: A+, 0 errors, 0 warnings for source and generated copies
- benchmark comparison: no regressions; total minimum 583.4 ms to 574.0 ms
- openshift-eng/ai-helpers downstream lint: exit 0, 0 errors, 0 warnings


skills/skillsaw-onboard/SKILL.md
stbenjam, line 1: I don't like the way this is worded because it forces the agent to basically read every file. We are missing a router here, so the file only needs to be read if the user opts into the specific action. For example, the link checker, you should have a section that says ask the user if they'd like to set up uh external link checking, blah blah blah, and then route to the file. Every file needs a routing decision. This avoids the bloat of having to load the whole skill into context.

not-stbenjam:
Addressed inline comment 3887632408: the onboarding entrypoint is now a decision router instead of an ordered runbook. Every reference has an explicit state or opt-in condition, references return to the router without forcing the next file, and external link checking is a separate opt-in Lychee reference. The source and compiled copies lint clean, and the structural test now verifies that every reference is routed exactly once.

.agents/skills/skillsaw-onboard/SKILL.md
```diff
@@ -10,324 +10,115 @@ metadata:
...
-config file. Tell the user they can configure rule settings in this file.
-
-If one already exists, keep it and skip this step.
+### 7. Add external link checking
 
-## Step 7: Set up CI
+If GitHub Actions is available, ask separately:
 
-Ask the user which CI system they use. Offer the following options:
+> I can add a separate weekly Lychee workflow that makes network requests to
```
stbenjam: Explain better - broken external links should be audited on a recurring basis, not on every lint - we reccomend using a link checker like Lychee. I can set that up for you, if you'd like?

not-stbenjam:
Addressed the supplied feedback at skills/skillsaw-onboard/SKILL.md line 93 in c7ac5740. The prompt now explains that external links need recurring audits, why network-dependent checks do not belong in every skillsaw lint, and why a dedicated checker such as Lychee is recommended before offering the weekly workflow setup.

PR #559, opened 2026-08-30
Simplify high-complexity audit hotspots

not-stbenjam:
## Summary

- reuse the repository scan for Promptfoo discovery and prune excluded eval trees
- distinguish suggest-only fixes from cases where no fixes exist
- extract APM attachment and centralize shared lint-tree build state
- decompose Copilot hook validation and OpenCode model lowering
- consolidate frontmatter rewriting behind one pure composition boundary

No rule defaults, severities, or enforcement semantics change in this PR.

## Architectural result

`build_lint_tree` drops from cyclomatic complexity 145 to 124 and cognitive complexity 485 to 402. The APM stage and shared mutable build state now have explicit boundaries. A broader plugin-loop extraction was deliberately deferred after adversarial review judged it too risky for the lint-tree core in this batch.

All seven commits were adversarially reviewed in isolation. Exact tree snapshots, node counts, violation ordering, and exit codes remained stable across representative APM, plugin, OpenCode, Copilot, and Promptfoo fixtures. Two independent roll-up reviewers approved the combined branch.

## Validation

- `make test`: 5,702 passed
- `make lint`: clean
- `make update`: clean; no generated diff
- medium benchmark: no regression; 1,373 nodes and 151 violations preserved
- Python 3.9 grammar checks passed for changed production files
- ai-helpers smoke test: branch and exact `origin/main` both report the same pre-existing stale `docs/index.html` custom-rule error; no branch-specific downstream delta

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved frontmatter updates, including empty, incomplete, and missing-frontmatter documents, while preserving file content.
  - Validation now reports all applicable hook issues instead of stopping after the first malformed field.
  - Improved model selection validation for supported string and mapping formats.
  - Promptfoo and related configuration discovery is more consistent across repository layouts and platforms.

- **CLI**
  - Clarified fix results with separate messages for suggested fixes and cases with no available fixes.

- **Tests**
  - Added coverage for frontmatter handling, validation diagnostics, repository discovery, tree output, and fix messaging.

### DIFF
--- a/src/skillsaw/blocks/frontmatter.py
+++ b/src/skillsaw/blocks/frontmatter.py
@@ -59,6 +59,34 @@ def _parse_file_frontmatter(
     return fm, None, None, body, fm_line_count
 
 
+def _compose_frontmatter_document(content: Optional[str], fm: str) -> str:
+    """Return *content* with its frontmatter replaced by normalized *fm*."""
+    if not content:
+        return f"---\n{fm}---\n"
+
+    match = _FRONTMATTER_RE.match(content)
+    if match:
+        return f"---\n{fm}---\n{content[match.end() :]}"
+
+    if not content.startswith("---"):
+        return f"---\n{fm}---\n{content}"
+
+    lines = content.split("\n")
+    close_idx = None
+    for index, line in enumerate(lines[1:], 1):
+        if line.strip() == "---":
+            close_idx = index
+            break
+    if close_idx is None:
+        body_after = "\n".join(lines[1:])
+        return f"---\n{fm}---\n{body_after}"
+
+    body_after = "\n".join(lines[close_idx + 1 :])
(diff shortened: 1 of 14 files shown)


not-stbenjam:
Rejecting the Codex suggestion in [discussion_r3890743451](https://github.com/stbenjam/skillsaw/pull/559#discussion_r3890743451). We reproduced the behavior, but it requires an undocumented and apparently unobserved layout where the supported `evals/` root is itself a directory symlink. Independent GitHub samples found no such usage, and conventionally named `promptfooconfig.yaml` files remain discoverable through the physical path. Preserving this incidental behavior would require broader YAML indexing or a special fallback traversal in the shared discovery core, undermining the one-walk simplification for no demonstrated common-user benefit. We can reconsider if a real-world repository reports the layout.

PR #609, opened 2026-09-12
[Auto] Accept referenced Codex manifest-directory assets

stbenjam:
The official OpenAI catalog now includes Temporal, Chatcut, Superpowers and Remotion plugins whose icons explicitly reference `.codex-plugin/assets/`. `codex-plugin-structure` reported four layout warnings despite those references naming existing files inside each plugin.

Accept explicitly referenced interface asset files in the manifest directory. Inspect their directories so unreferenced siblings remain visible, and require the declared scalar/list field shape before granting an exemption. Unreferenced entries and missing or escaping references retain their diagnostics. Share the asset field vocabulary with manifest validation, correct the rule documentation, and add fixture-based CLI and unit regression coverage.

Validation: `make test` (9,278 passed, one skipped), `make lint`, `make update`, and `skillsaw lint` on `openshift-eng/ai-helpers` all passed. The six Codex rules now report zero findings on `openai/plugins` (62 plugins, 507 skills), including under `--strict`.

Sources: [official catalog](https://github.com/openai/plugins), [plugin field guide](https://github.com/openai/codex/blob/rust-v0.154.0/codex-rs/skills/src/assets/samples/plugin-creator/references/plugin-json-spec.md). The prose documentation still recommends root-level assets; this exception follows explicitly referenced files in the official conformance corpus.


stbenjam:
No, don't suggest this -- it is demonstrably wrong according to the spec.

PR #616, opened 2026-09-19
[Auto] Accept Unicode names in Agent Skills

not-stbenjam:
The Agent Skills specification permits Unicode lowercase alphanumerics, but `agentskill-name` rejected valid names such as `café` and `数据分析`. Its rename helper could also strip non-Latin letters from a suggested name. Accept Unicode letters and numbers and preserve them during autofixes, while retaining lowercase, hyphen, length, and directory-match checks.

The governing requirement is the [published name specification](https://agentskills.io/specification#name-field), corroborated by the [reference validator](https://github.com/agentskills/agentskills/blob/547831f3a23724ba64a9b79bbf59c5e0bc8f2d1a/skills-ref/src/skills_ref/validator.py). Literal directory matching remains unchanged; the reference validator's additional NFKC normalization is not specified by the prose requirement.

Regression coverage includes Latin accents, Cyrillic, Chinese, Arabic digits, invalid characters, unchanged valid files after CLI fix, Unicode-preserving renames, and SAFE-autofix convergence. The fixture tracking check now uses NUL-delimited Git output so Unicode paths are compared literally. Updated the rule documentation and maintenance reference.

Validation: `make test` (9,338 passed), `make lint`, `make update`, scoped context lint, and default lint of a fresh `openshift-eng/ai-helpers` checkout (exit 0, no errors or warnings). The official OpenAI plugin corpus retains four documented `.codex-plugin/assets` layout findings; no check was weakened for those entries.

CI also exposed a stale Docker action version annotation on upstream main: the moving `v7` tag now names 7.4.0, while the pinned SHA is 7.3.0. Corrected the comment to `v7.3.0` without changing executable code. Reproduced the original failure with zizmor 1.29.0 online and verified the correction passes.


stbenjam:
What? The spec literally says: `May only contain unicode lowercase alphanumeric characters (a-z, 0-9) and hyphens (-)`

not-stbenjam:
Withdrawing this change. I treated the word “unicode” as permission to broaden the character set despite the explicit a-z, 0-9 range in the published specification, then relied on the more permissive reference validator. This is conflicting evidence, not a confirmed specification change; the maintenance workflow requires preserving the existing check in that situation. The independent Docker action version-comment correction is also included in #617.

